PR #54, opened 2026-03-19
fix: normalize quoted user-invokable frontmatter

Gujiassh:
## Summary
- normalize quoted `user-invokable` frontmatter booleans so `'true'`/`"true"` are treated the same as bare `true`
- keep quoted booleans on other frontmatter keys as strings instead of broadening parser behavior
- add focused regressions for both `parseFrontmatter()` and `readSourceFiles()` on quoted user-invokable values

## Validation
- `bun test tests/lib/utils.test.js`
- `bun test`

### DIFF
--- a/scripts/lib/utils.js
+++ b/scripts/lib/utils.js
@@ -59,9 +59,13 @@ export function parseFrontmatter(content) {
       if (colonIndex > 0) {
         const key = trimmed.slice(0, colonIndex).trim();
         const value = trimmed.slice(colonIndex + 1).trim();
+        const normalizedValue =
+          key === 'user-invokable' && /^['"](?:true|false)['"]$/.test(value)
+            ? value.slice(1, -1)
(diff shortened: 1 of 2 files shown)


pbakaus (changes requested):
Thanks for the contribution! A couple of issues:

1. **Wrong key name**: The frontmatter key is `user-invocable` (with a **c**), not `user-invokable` (with a **k**). We fixed this across the codebase in PR #50. All references in this PR use the wrong spelling.

2. **Not sure this fix is needed**: The source SKILL.md files are maintained by this project -- we control the frontmatter and always write `user-invocable: true` as a bare boolean, never quoted. I don't see a real-world scenario where quoted booleans (`'true'` / `"true"`) would appear in these files.

If you have an actual use case where this is causing a problem, I'd love to hear about it. Otherwise I'm going to close this.

PR #64, opened 2026-03-23
feat: add symlink script for Git submodule usage

### LINKED ISSUE #46: Git submodule support
## What problem does this solve?

I manage my projects via git and don't want to worry about running additional shell commands to update or download a zip file manually each time.

## Proposed solution

Offer a project structure or branch that can be directly included as a git submodule and further handled via git update commands.

## Provider(s) this applies to

- [ ] Cursor
- [ ] Claude Code
- [ ] Gemini CLI
- [ ] Codex CLI
- [ ] VS Code Copilot
- [ ] Kiro
- [ ] OpenCode
- [X] All providers

## Alternatives considered

Include an update script in the universal zip file.

amirgeek:
Fixes #46. 

This PR adds a convenient shell script (`bin/link.sh`) and documentation for using the repository as a git submodule. This allows users to keep their skills up to date with `git submodule update --remote` without having to manually download ZIPs or run update scripts.

The script automatically detects the provider you want and creates a relative symlink from your project's root to the submodule directory.

## Usage:
```bash
git submodule add https://github.com/pbakaus/impeccable .impeccable
./.impeccable/bin/link.sh cursor
```

### DIFF
--- a/bin/link.sh
+++ b/bin/link.sh
@@ -19,20 +19,47 @@ if [ -z "$PROVIDER" ]; then
 fi
 
 PROVIDER_DIR=".$PROVIDER"
+SOURCE_SKILLS_DIR="$REPO_DIR/$PROVIDER_DIR/skills"
+TARGET_PROVIDER_DIR="$PARENT_DIR/$PROVIDER_DIR"
+TARGET_SKILLS_DIR="$TARGET_PROVIDER_DIR/skills"
 
 if [ ! -d "$REPO_DIR/$PROVIDER_DIR" ]; then
   echo "Error: Provider directory '$PROVIDER_DIR' not found in $REPO_DIR"
   exit 1
 fi
 
-echo "Linking .$PROVIDER to parent directory..."
-
-# Check if target already exists in parent
-if [ -e "$PARENT_DIR/$PROVIDER_DIR" ]; then
-  echo "Warning: '$PARENT_DIR/$PROVIDER_DIR' already exists. Skipping."
-else
-  # Create relative symlink in the parent directory
-  cd "$PARENT_DIR"
-  ln -s "$SUBMODULE_NAME/$PROVIDER_DIR" "$PROVIDER_DIR"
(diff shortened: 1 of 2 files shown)


codeCraft-Ritik (commented):
Excellent improvement

bin/link.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if [ ! -d "$REPO_DIR/$PROVIDER_DIR" ]; then
+  echo "Error: Provider directory '$PROVIDER_DIR' not found in $REPO_DIR"
+  exit 1
+fi
+
+echo "Linking .$PROVIDER to parent directory..."
+
+# Check if target already exists in parent
+if [ -e "$PARENT_DIR/$PROVIDER_DIR" ]; then
```
pbakaus: I'm assuming that a lot of people will already have e.g. a .claude folder, so this technique wouldn't work then...would it not be better to symlink the sub folders (.claude/skills/{folders in here})?

pbakaus (changes requested):
Thanks for your contribution! I have one suggestion that I think would make this much easier for adoption, otherwise looks promising.

abdulwahabone:
> Good catch — updated the script so it now creates symlinks only inside the provider's skills folder (e.g. `.claude/skills/*`) instead of linking the whole provider directory. That way existing `.claude` / `.cursor` setups stay intact. I also updated the README to clarify the new behavior.

Hey @pbakaus @amirgeek ran into the same issue on the npx impeccable skills install path and put up a fix in #148 that takes the same approach. Two sides of the same problem, might be worth having a look.

pbakaus:
Thanks for putting this together. The core idea was right: submodule users need a way to link only the provider skill folders, not entire harness config directories.

I’m going to close this PR because it’s now stale/conflicting and the implementation has been superseded by a CLI-native version. The new path is:

npx impeccable skills link --source=.impeccable --providers=claude,cursor

That reuses the existing installer’s provider resolution, links from the compiled `dist/universal` output, handles provider aliases, avoids overwriting real skill directories unless `--force` is passed, and keeps `skills update` from replacing symlinked installs.

Appreciate the original push here. It shaped the final direction.

PR #65, opened 2026-03-23
fix: show $ instead of / for Codex command references

gabelul:
This PR isolates the Codex command-prefix fix from #35.

## What changed

- make Codex-generated command references use `$` instead of `/`
- thread the Codex command prefix through placeholder replacement
- make Codex-prefixed skill references use the Codex invocation style consistently

## Why

Codex commands are invoked with `$`, so showing `/normalize` or `/audit` in generated Codex skill content is misleading. This keeps the generated Codex output aligned with actual Codex usage.

## Scope

This PR intentionally includes only the Codex prefix fix.

It does not include the `/next` / `/fix-all` recommendation workflow from #35.

## Validation

- targeted tests passed:
  - `tests/lib/utils.test.js`
  - `tests/lib/transformers/codex.test.js`
  - `tests/build.test.js`

This branch was rebuilt from current `main` and narrowed to just this fix.

### DIFF
--- a/scripts/lib/transformers/codex.js
+++ b/scripts/lib/transformers/codex.js
@@ -46,8 +46,8 @@ export function transformCodex(skills, distDir, patterns = null, options = {}) {
 
     const frontmatter = generateYamlFrontmatter(frontmatterObj);
 
-    let skillBody = replacePlaceholders(skill.body, 'codex', commandNames);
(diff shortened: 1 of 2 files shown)


pbakaus (approved):
Looks great, thank you for your contribution! Much appreciated.

PR #87, opened 2026-04-05
fix: normalize quoted user-invocable frontmatter

Gujiassh:
## Summary
- normalize quoted boolean frontmatter values only for the `user-invocable` flag (while keeping the legacy `user-invokable` spelling readable too)
- stop coercing unrelated quoted values like `description: 'true'` into booleans
- add focused regression coverage for quoted flags in both raw frontmatter parsing and `readSourceFiles`

## Testing
- `bun test tests/lib/utils.test.js`
- `PATH=/home/cc/tmp/bunshim:$PATH bun run build`

### DIFF
--- a/scripts/lib/utils.js
+++ b/scripts/lib/utils.js
@@ -62,13 +62,19 @@ export function parseFrontmatter(content) {
       if (colonIndex > 0) {
         const key = trimmed.slice(0, colonIndex).trim();
         const value = trimmed.slice(colonIndex + 1).trim();
-        const normalizedValue =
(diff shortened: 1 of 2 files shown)


Gujiassh:
Refresh update: rebased this branch onto latest origin/main and force-pushed.

Validation run:
- bun test tests/lib/utils.test.js

Result:
- pass (62 passed, 0 failed).

pbakaus (approved):
Correct YAML-semantics fix, with good regression coverage on both `parseFrontmatter` and `readSourceFiles`. Thanks for sticking with this through the rebases!

PR #90, opened 2026-04-07
fix: add trailing slash to skills path in plugin.json

octo-patch:
Fixes #86

## Problem
The `skills` path in `.claude-plugin/plugin.json` was missing a trailing slash, which caused Claude Code to fail to recognize slash commands after installing the plugin.

## Solution
Added the missing trailing slash to the `skills` field:
```
-  "skills": "./.claude/skills"
+  "skills": "./.claude/skills/"
```

## Testing
All 220 existing tests pass (`bun test`). The fix was originally identified by the issue reporter, and this PR applies that exact one-character change.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -8,5 +8,5 @@
   },
   "homepage": "https://impeccable.style",
   "repository": "https://github.com/pbakaus/impeccable",
-  "skills": "./.claude/skills"
+  "skills": "./.claude/skills/"
 }


pbakaus:
I've tested this with and without slash and could not see any difference. Please retest with the latest version (only a single skill). If it still is an issue, feel free to let me know and I'll reopen.

pbakaus:
Actually - I stand corrected. Looks like the docs have been updated. I still couldn't confirm, but to be safe, I'm adding a slash for v3.0.2. Thanks for your report!

PR #96, opened 2026-04-11
fix(windows): resolve browser script paths from file URLs

### LINKED ISSUE #95: Windows: detect command fails with doubled drive letter in path
## Bug

`npx impeccable detect <url>` fails on Windows with:

```
Error: Browser script not found at C:\C:\Users\...\detect-antipatterns-browser.js
```

Note the doubled `C:\C:\`.

## Cause

In `src/detect-antipatterns.mjs`, two locations construct the path to `detect-antipatterns-browser.js` using:

```js
path.dirname(new URL(import.meta.url).pathname)
```

On Windows, `new URL('file:///C:/foo/bar').pathname` returns `/C:/foo/bar` (with a leading slash). When passed to `path.join` or `path.resolve`, Node prepends the drive letter again, producing `C:\C:\...`.

**Affected lines:**
- ~L2690: `path.dirname(new URL(import.meta.url).pathname)` (puppeteer scan path)
- ~L3506: `path.dirname(new URL(import.meta.url).pathname)` (live detect path)

## Fix

Replace `new URL(import.meta.url).pathname` with `fileURLToPath(import.meta.url)` from `node:url`, which correctly handles Windows drive letters:

```js
import { fileURLToPath } from 'node:url';
const scriptPath = path.join(path.dirname(fileURLToPath(import.meta.url)), 'detect-antipatterns-browser.js');
```

## Environment

- Windows 11 Pro
- Node.js (via npm global install)
- impeccable installed via `npm i -g impeccable`

Gujiassh:
## Summary

- replace `new URL(import.meta.url).pathname` with a `fileURLToPath`-based helper when resolving `detect-antipatterns-browser.js`
- reuse the same helper for both URL scanning and live server injection paths
- add a Windows path regression test so drive letters do not get doubled

## Testing

- `node - <<'NODE' ... browserScriptPathFromModulePath(..., path.win32) ... NODE`
- `bun test tests/detect-antipatterns.test.js`
  - the new Windows path regression passes
  - the suite still has the pre-existing `linked stylesheet detected (jsdom default)` failure unrelated to this change

Fixes #95

### DIFF
--- a/src/detect-antipatterns.mjs
+++ b/src/detect-antipatterns.mjs
@@ -26,10 +26,11 @@ const IS_BROWSER = typeof window !== 'undefined';
 const IS_NODE = !IS_BROWSER;
 
 // @browser-strip-start
-let fs, path;
+let fs, path, fileURLToPath;
 if (!IS_BROWSER) {
   fs = (await import('node:fs')).default;
   path = (await import('node:path')).default;
+  ({ fileURLToPath } = await import('node:url'));
 }
 // @browser-strip-end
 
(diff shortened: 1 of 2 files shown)


pbakaus:
Closing as superseded. The same bug (#95) was fixed by #101 (`fix: use fileURLToPath for Windows path resolution`) plus a follow-up that added the Windows regression test at `tests/windows-path-fix.test.js`. Thanks for the patch, sorry the timing didn't work out.

PR #98, opened 2026-04-12
fix(plugin): add trailing slash to skills path

Gujiassh:
## Summary\n- update .claude-plugin/plugin.json to use ./\.claude/skills/ for the skills path\n- add a regression test asserting plugin skills path stays normalized with a trailing slash\n\n## Validation\n- node --test tests/plugin-json.test.mjs\n\nFixes #86

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -8,5 +8,5 @@
   },
   "homepage": "https://impeccable.style",
   "repository": "https://github.com/pbakaus/impeccable",
-  "skills": "./.claude/skills"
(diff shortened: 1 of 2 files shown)


pbakaus:
I've tested this with and without slash and could not see any difference. Please retest with the latest version (only a single skill). If it still is an issue, feel free to let me know and I'll reopen.

pbakaus:
Actually - I stand corrected. Looks like the docs have been updated. I still couldn't confirm, but to be safe, I'm adding a slash for v3.0.2. Thanks for your report!

PR #123, opened 2026-04-29
feat(load-context): support context dir outside repo root (closes #119)

### LINKED ISSUE #119: [Feature] Have the option to keep `DESIGN.md` and `PRODUCT.md` in a `docs/` folder.
## What problem does this solve?

I hate having lots of .md files in my project root, so I tuck away as many of them as I can into a `docs/` folder.

## Proposed solution

Right now all the impeccable skills look for `DESIGN.md`, `DESIGN.json` and `PRODUCT.md` in the root folder. It would be good to have a fallback where they can look them up in another folder like `docs/`.

## Provider(s) this applies to

- [ ] Cursor
- [ ] Claude Code
- [ ] Gemini CLI
- [ ] Codex CLI
- [ ] VS Code Copilot
- [ ] Kiro
- [ ] OpenCode
- [x] All providers

## Alternatives considered

none?

0x963D:
## What

Adds a configurable lookup path for `PRODUCT.md` / `DESIGN.md` / `DESIGN.json` so the three impeccable docs don't have to live at the project root.

Resolution order, first match wins:

1. `process.env.IMPECCABLE_CONTEXT_DIR` — absolute or relative to cwd
2. `cwd`, when canonical or legacy files are at the root (back-compat)
3. Auto-fallback subdirs of cwd: `.agents/context/` then `docs/`
4. `cwd` as a default "no context found" location (preserves the current `hasProduct === false` semantics)

Closes #119.

## Why

Current loader hard-codes `process.cwd()`. Users who keep their root tidy (`AGENTS.md` auto-imports under `.agents/context/`, or a `docs/` folder per the issue) have to either litter the root with three big spec files or fork/patch the skill. Forking gets clobbered on every skill sync because the scripts are tracked by hash in `skills-lock.json`.

Step 2 keeps the existing root-level layout working unchanged. Step 3 covers the two conventions actually seen in the wild — `.agents/context/` for `AGENTS.md`-based auto-import setups, and `docs/` for the request in #119 — without any configuration. Step 1 is the escape hatch for anything else.

## Changes

- `source/skills/impeccable/scripts/load-context.mjs` — export `resolveContextDir()` and use it inside `loadContext()`; add `contextDir` to the JSON output.
- `source/skills/impeccable/scripts/live-server.mjs` — import `resolveContextDir` and read `PRODUCT.md` / `DESIGN.md` / `DESIGN.json` from the resolved dir instead of `process.cwd()` directly. (Three call-sites: `hasProjectContext()` and the `/design-system.json` handler.)
- `source/skills/impeccable/SKILL.md` — short note in **Setup → Context gathering** explaining the env var and fallback dirs.
- `tests/load-context.test.mjs` — new file, 19 cases under `node --test`. Covers env-var override (relative + absolute + empty), all four resolution steps, fallback precedence (`.agents/context/` over `docs/`, root over both), the legacy `.impeccable.md` migration staying scoped to cwd root, and the existing back-compat shape of `loadContext()`.
- Provider sync via `bun run build` (the 13 derivative dirs + `plugin/` follow `source/`).

## Notes on scope

- Legacy `.impeccable.md` → `PRODUCT.md` auto-migration only fires at `cwd` root. Fallback dirs are read-only as far as auto-rename is concerned — we don't surprise users by mutating files inside `docs/` or `.agents/`.
- `productPath` / `designPath` in the JSON output remain relative to `cwd` (existing contract). The new `contextDir` field is the absolute resolved dir.
- No new dependencies. No public API removed.

## Tests

- `node --test tests/load-context.test.mjs` — 19/19 pass on the new file.
- `node --test tests/live-server.test.mjs` — 22/22 pass (no regressions from the live-server swap).
- `bun run test` — 169/170 pass on this branch. The single failure (`tests/windows-path-fix.test.js > fileURLToPath handles POSIX file URLs correctly`) reproduces identically on `upstream/main` on Windows and is unrelated to this PR.
- `bun run build` clean.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 8775b01fd980ac7f763d49b98f66854a3254434f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/skills/impeccable/SKILL.md
+++ b/.agents/skills/impeccable/SKILL.md
@@ -30,7 +30,7 @@ Other harnesses should follow the same checklist when they can expose this state
 
 ### 1. Context gathering
 
-Two files at the project root, case-insensitive:
+Two files, case-insensitive. The loader looks at the project root by default and falls back to `.agents/context/` and `docs/` if the root is clean. Override with `IMPECCABLE_CONTEXT_DIR=path/to/dir` (absolute or relative to cwd).
 
 - **PRODUCT.md** — required. Users, brand, tone, anti-references, strategic principles.
 - **DESIGN.md** — optional, strongly recommended. Colors, typography, elevation, components.
@@ -41,7 +41,7 @@ Load both in one call:
 node .agents/skills/impeccable/scripts/load-context.mjs
 ```
 
-Consume the full JSON output. Never pipe through `head`, `tail`, `grep`, or `jq`.
+Consume the full JSON output. Never pipe through `head`, `tail`, `grep`, or `jq`. The output's `contextDir` field tells you where the files were resolved from.
 
 If the output is already in this session's conversation history, don't re-run. Exceptions requiring a fresh load: you just ran `$impeccable teach` or `$impeccable document` (they rewrite the files), or the user manually edited one.
 
--- a/.agents/skills/impeccable/scripts/live-server.mjs
+++ b/.agents/skills/impeccable/scripts/live-server.mjs
@@ -21,11 +21,16 @@ import path from 'node:path';
 import net from 'node:net';
 import { fileURLToPath } from 'node:url';
 import { parseDesignMd } from './design-parser.mjs';
+import { resolveContextDir } from './load-context.mjs';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 // PID file in the project root so both the server and agent can find it
 // predictably (os.tmpdir() varies across platforms).
 const LIVE_PID_FILE = path.join(process.cwd(), '.impeccable-live.json');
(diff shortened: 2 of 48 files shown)


pbakaus (approved):
Excellent work. Closes a real user request, preserves backward compat with cwd-first resolution, includes proper escape hatches (env var + auto-fallback), and the test coverage is thorough (19 cases on isolated scratch dirs). The auto-migration scoping (legacy `.impeccable.md` only renamed at cwd root, not inside fallback dirs) is exactly the right call. Verified locally: clean merge against current main, all 186 tests pass, build clean across all 13 mirrors. Thanks for sending this in!

PR #160, opened 2026-05-18
fix: add mandatory mechanical pre-scan to typeset and layout commands

### LINKED ISSUE #149: [Bug] Typeset and Layout often need to be run mutliple times to catch basic things
## What happened?

When running a /typeset or /layout on a new design imported from Claude design, I find that I'm needing to get sub-agents to double-check their work quite a bit. They're missing basic elements that could be caught in a mechanical check. I'm putting this in as a bug - as it's unexpected behaviour. 

## Steps to reproduce

1. Run /tpeset or /layout on a design from Claude Design 
2. Get a subagent to check the original agent's work- find text etc that is off ramp. 
3. 

## Expected behavior
In the first or the second run, an agent would catch things like 12.5px font size. 

## Provider & environment

- **Provider** (Cursor / Claude Code / Gemini CLI / Codex / Copilot / Kiro / OpenCode): Claude Code Opus 4.7 Extra High 
- **Provider version**: Claude 1.6608.2
- **OS**:  MacOS 15.6

## Additional context

I've been through this loop a few times on a few different files, this time an agent wrote a memory that could be useful context to add to skills

---
name: Mechanical scan before visual review on cleanup passes
description: On typeset / theming / token-cleanup passes, run a regex/grep scan for off-spec values before relying on visual review.
type: feedback
---

On typeset / theming / token-cleanup passes, run a mechanical scan (grep/regex) for off-spec values *before* relying on visual review. Examples:
- Orphan font sizes: grep for `text-\[\d+(\.\d+)?px\]` and verify each result against the documented ramp.
- Ad-hoc line-heights: grep for `leading-\[` and check each value against the per-role spec.
- Hard-coded colours: grep for `#[0-9a-fA-F]{3,6}` in `.tsx` / `.css` to surface untoken'd hexes.
- Orphan opacity values, orphan spacing, etc. — same approach.

**Why:** On the SearchResults typeset pass, three off-grid values slipped past visual review (filter-pill count badge at 10px, missing `leading-none` on Label-tier elements, chip line-height not matching the documented exception). All three were findable with a single grep in seconds. The audit agent caught them after the fact; the cleanup agent should catch them up front.

**How to apply:** Before declaring a typeset/theming/token pass complete, run the relevant greps and verify every result against the documented spec. Surface remaining off-spec values in the response so the user can decide which to fix vs accept as scoped exceptions.

kaushalrog:
Resolves #149. Agents running /typeset or /layout now grep for
off-spec values (orphan px font sizes, ad-hoc leading, hard-coded
hex colors, arbitrary spacing) before visual review. This catches
values like 12.5px that slip past LLM visual passes.

Surfaces all hits in the response so the user can decide which to
fix vs. accept as scoped exceptions.

---

> [!NOTE]
> **Low Risk**
> Low risk documentation-only change that adds required pre-review grep checks; the main risk is minor confusion if the regexes or the stray trailing `\` in `layout.md` are incorrect.
> 
> **Overview**
> Adds a **mandatory “Mechanical Pre-Scan” step** to the `layout` and `typeset` reference guides, requiring reviewers to run predefined `grep` commands before visual analysis.
> 
> The new pre-scan checks surface off-spec values (e.g., Tailwind arbitrary spacing/typography, orphan `opacity`/`leading`, hard-coded `px` spacing/font sizes, hex colors, and extreme `z-index` values) and instructs agents to report all matches so users can choose fixes vs. scoped exceptions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e4f27cd1d5b5606018afa276ba242be12af49bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skill/reference/layout.md
+++ b/skill/reference/layout.md
@@ -1,6 +1,23 @@
 Space is the most underused design tool. Find the layout's actual problem (monotone spacing, weak hierarchy, identical card grids, the centered-stack default) and fix the structure, not the surface.
 
 ---
+## Mechanical Pre-Scan (required before visual review)
+
+Run these greps on the target files before any visual analysis. Surface every hit in your response so the user can decide which to fix vs. accept as a scoped exception.
+
+
+# Arbitrary spacing values outside the scale (Tailwind arbitrary syntax)
+grep -rn '\(p\|m\|gap\|space\)-\[[0-9]*\.?[0-9]*\(px\|rem\|em\)\]' --include="*.tsx" --include="*.ts" --include="*.css" --include="*.jsx" --include="*.js" .
+
+# Orphan opacity values
+grep -rn 'opacity-\[' --include="*.tsx" --include="*.ts" --include="*.css" --include="*.jsx" --include="*.js" .
+
+# Hard-coded pixel spacing in raw CSS
+grep -rn '\(padding\|margin\|gap\):\s*[0-9]*\.?[0-9]*px' --include="*.css" --include="*.scss" .
+
+# Arbitrary z-index values (999, 9999, etc.)
+grep -rn 'z-\[9\{2,\}\]' --include="*.tsx" --include="*.ts" --include="*.css" --include="*.jsx" --include="*.js" .
+\
 
 ## Register
 
--- a/skill/reference/typeset.md
+++ b/skill/reference/typeset.md
@@ -1,6 +1,21 @@
 Typography carries most of the information on the page. Replace generic defaults (Inter, Roboto, system fallback at flat scale) with type that reflects the brand and scales with intentional contrast.
 
 ---
+## Mechanical Pre-Scan (required before visual review)
+
+Run these greps on the target files before any visual analysis. Surface every hit in your response so the user can decide which to fix vs. accept as a scoped exception.
+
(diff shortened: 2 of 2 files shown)


Nystamere (approved):

Nystamere (approved):

Nystamere (approved):

pbakaus:
Thanks for this. I agree with the intent: `/typeset` and `/layout` need a mechanical preflight so we catch obvious value-level issues before asking the model to do a higher-level design pass.

I don’t think this particular implementation is the right shape to merge, though. The grep commands are brittle, currently have a few regex/shell issues, and are mostly Tailwind-arbitrary-value specific, so they’d miss a lot of the ways these problems show up in real projects.

I think this should move into the detector instead: add a typography-scoped preflight, reuse/extend the existing typography-related detector rules, and have `/typeset` run that first before doing the full LLM typography review. That also lines up better with #149 and should compose cleanly with the hook work in #170.

I’m going to close this PR in favor of that detector-backed approach, but the underlying idea is definitely valid.

PR #203, opened 2026-06-04
Add regression guard for light-mode periodic table

### LINKED ISSUE #201: [Bug] Light mode periodic table becomes unreadable
## What happened?

I visited the website on light mode and when I checked the periodic table I couldn't read any skill because the text had become black on black. A project that is all about design should be able to catch stuff like this. Perhaps an extra skill is needed?

## Steps to reproduce

1. Visit the website
2. Turn on Light Mode
3. Click on periodic

## Expected behavior

I expect that the design is correctly adapted to light mode so I can read the elements/skills.

## Additional context

Image

abdulwahabone:
## Summary
- Add a homepage light-mode regression test for the periodic table view.
- Lock the real emitted tile class, `.ptable-element`, to the light-mode override so the old stale-selector bug cannot return.
- Assert light-mode periodic tiles use light surface tokens with dark text tokens, and include the test in `bun run test`.

## Repro / verification notes
- Reproduced the original issue on `50830001^` in Light Mode + Periodic: tiles stayed `oklch(0.15 0 0)` while text resolved to dark light-mode tokens.
- Verified this branch/current main renders the same view with tile backgrounds at `oklch(0.99 0.008 95)` and readable dark text.
- Fresh main already contains the visual CSS fix from `50830001`; this PR adds the missing regression guard.

## Screenshots

Before fix, Light Mode + Periodic:

Before fix: unreadable dark tiles in light mode

After fix, Light Mode + Periodic:

After fix: readable light tiles in light mode

Closes #201

## Validation
- `bun test tests/site-light-mode.test.js`
- `bun test tests/build.test.js tests/site-light-mode.test.js`
- `bun run build`
- `bun run test`


abdulwahabone:
Closing this draft: fresh `main` already includes the visual fix in `50830001` (`Fix periodic-table view in light mode`). I verified the bug only reproduces on the parent commit and current `main` renders the light-mode periodic table correctly, so we do not need this extra regression-test PR.

PR #217, opened 2026-06-07
Code health: centralize thresholds, finding provenance, types, DRY border rule

cameron-creations:
Four low-risk code-health improvements. **No change to which anti-patterns fire** (detect suites unchanged: 172 bun + 48 node).

### 1. Centralize numeric thresholds (`cli/engine/shared/thresholds.mjs`)
Pulls scattered magic numbers into one documented module: the 50-file warning cutoff, the 500ms TCP / 2000ms HTTP dev-server probe timeouts, and the default `lineLengthMax` of 80. Wired into `cli/main.mjs`, `node/file-system.mjs`, and `rules/checks.mjs`.

### 2. Finding provenance (`engine` + `category`)
`finding()` now stamps `category` (from the registry) and each engine tags its findings with `engine` (`regex` / `static-html` / `browser`). Text output shows `[rule] (category/engine)`; `--json` carries the new fields additively (shape unchanged otherwise). Helps users judge confidence and where a finding came from.

### 3. Public type declarations (`cli/engine/index.d.ts`)
Hand-written `.d.ts` for `detectText` / `detectHtml` / `detectUrl` / `formatFindings` plus `Finding`, `ScanOptions`, and `ImpeccableConfig` types, wired through `package.json` `exports.types`. Validated with `tsc --strict`.

### 4. DRY the side-tab border thresholds
The static-HTML and regex engines both hard-coded the strong-accent border width (`>= 3px`). Extracted to `BORDER_ACCENT_STRONG_WIDTH_PX` / `BORDER_ACCENT_MIN_WIDTH_PX` and shared. Note: the regex engine intentionally keeps its own coarser radius-aware variants (`>= 1` rounded, `>= 4` shorthand) since it lacks layout context; only the genuinely-shared `3`/`2` values were unified, not the divergent ones.

No new dependencies. New tests added to the `test` script.

---

> [!NOTE]
> **Low Risk**
> Refactor and additive API/output fields only; threshold values and rule logic are preserved, with new tests locking defaults and provenance.
> 
> **Overview**
> This PR is a **code-health** pass: it does not change which rules fire (same numeric behavior, documented as unchanged detect suites).
> 
[...]

PR #218, opened 2026-06-07
Add project config + rule suppression (config file + inline directives)

cameron-creations:
Adds a CLI project config and inline rule suppression. No new dependencies; detect suites unchanged (175 bun + 48 node, the +3 are new e2e tests).

> Note: this branch is stacked on #217 (code-health). The first four commits here are from that PR; review/merge #217 first and this diff reduces to just the config + suppression work. The runtime code does not depend on #217.

### Config file
Discovered from the scan target upward, first match wins: `impeccable.config.json` -> `.impeccablerc.json` -> `package.json#impeccable`. JSON only (no executable config), to preserve the no-code-execution posture.

```json
{
  "disabledRules": ["side-tab"],
  "ignore": ["vendor/**", "**/*.min.css"],
  "severity": { "overused-font": "off", "low-contrast-text": "error" },
  "lineLengthMax": 100
}
```

- `disabledRules`: drop rules from results.
- `ignore`: dependency-free glob matcher (`*`, `**`, `?`) on paths relative to the scan dir; matches are skipped before scanning.
- `severity`: override reported level (`note`/`warning`/`error`) or `off`.
- `lineLengthMax`: threads into the long-line check.

### Inline suppression
`impeccable-disable-line`, `impeccable-disable-next-line`, and file-level `impeccable-disable`, each with an optional rule-id list (omit to suppress all). Works in `//`, `/* */`, and `<!-- -->` comments.

### Notes
- Config resolves from the scan target's directory, not `process.cwd()`.
- New modules: `cli/engine/config/{load-config,glob,suppress}.mjs`, each unit-tested, plus a CLI e2e test. All registered in the `test` script. README documents the feature.

---

> [!NOTE]
> **Medium Risk**
> Changes what findings users see (config/suppression) and which files are scanned (ignore globs); behavior is well-tested but misconfiguration could hide real issues silently.
> 
> **Overview**
> Adds **project configuration** and **inline rule suppression** to the `impeccable detect` CLI, documented in README.
> 
> **Config** is discovered by walking upward from the **scan target** (not `process.cwd()`): `impeccable.config.json` → `.impeccablerc.json` → `package.json#impeccable`. JSON-only. Options include `disabledRules`, `ignore` (globs skip files before scanning), `severity` overrides/`off`, and `lineLengthMax` passed into scans.
> 
> After detection, the CLI filters findings through config and **`impeccable-disable-line` / `-next-line` / file-level `impeccable-disable`** comments (parsed from already-read source). New modules: `cli/engine/config/{load-config,glob,suppress}.mjs`.
> 
> **Related polish in the same diff:** findings gain `category` and `engine` (human output shows `category/engine`); shared `thresholds.mjs`; hand-maintained `index.d.ts` and package `types` export; `loadConfig` re-exported from the public API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2847ec4fb853871ad16b75c9f6c8cd0d95f41c88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pbakaus (changes requested):
Thanks for putting this together. I like the feature direction: project-level detector configuration and inline suppressions are both useful.

I don’t think this should merge as-is though:

The biggest issue is that much of the config work now overlaps with the newer unified detector config that already exists on `main`:

- `.impeccable/config.json` / `.impeccable/config.local.json`
- `detector.ignoreRules`
- `detector.ignoreFiles`
- `detector.ignoreValues`
- `detector.designSystem.enabled`
- `npx impeccable ignores ...`
- `npx impeccable detect --no-config`
- shared behavior between manual `detect` and the design hook

So this PR’s config format would introduce a second, incompatible system:

- `disabledRules` overlaps with `detector.ignoreRules`
- `ignore` overlaps with `detector.ignoreFiles`
- the new glob matcher overlaps with existing config glob helpers
- `impeccable.config.json`, `.impeccablerc.json`, and `package.json#impeccable` would sit beside the current `.impeccable/config*.json` model

Separately, the branch is stale/conflicted against current `main`, and it is still stacked on #217.

The parts that still seem worth salvaging are:

- inline suppression comments
- severity overrides, probably under the existing `detector` config namespace
- maybe `lineLengthMax`, also under `detector`, with clear docs about which scan paths it affects
- target-upward config discovery, if we decide current `process.cwd()` behavior is insufficient

The inline suppression implementation also needs another pass. I reproduced a few gaps:

1. HTML/static findings often have `line: 0`, so directives like this do not suppress the finding:

```html
<!-- impeccable-disable-next-line side-tab -->
<div style="border-left:4px solid red;border-radius:8px">x</div>
```
2. File-level disables have the same issue for lineless findings like overused-font.
3. Stdin scans skip inline suppression because the stdin source is never stored for the suppression filter.

The focused tests on the branch pass, and the config/ignore/severity direction is reasonable. But I’d suggest rebasing onto current main, dropping the parallel config loader/glob/config-file format, reusing the existing .impeccable/config*.json helpers, and keeping the PR focused on the missing capabilities.

PR #219, opened 2026-06-07
CI integration: SARIF output + reusable GitHub Action

cameron-creations:
Makes Impeccable usable as a CI gate. No new dependencies; detect suites unchanged (the +2 bun are new SARIF e2e tests, verified against `demos/`: 20 results, 7 rules, valid SARIF 2.1.0).

> Note: stacked on #217 (code-health). It reuses the `category`/`engine` provenance from that PR in the SARIF rule metadata. Review/merge #217 first and this diff reduces to just the SARIF + Action work.

### `--sarif`
New flag emits a SARIF 2.1.0 document to stdout (`cli/engine/output/sarif.mjs`). Severity maps to SARIF levels (`advisory` -> `note`, `warning` -> `warning`), paths are relativized, rules are de-duplicated into `tool.driver.rules`. SARIF mode exits 0 even with findings so a downstream upload step always runs. `--json` is unchanged.

### Reusable Action (`action.yml`)
Composite action that runs the scan and writes an uploadable SARIF file:

```yaml
- uses: pbakaus/impeccable@v2
  id: impeccable
  with: { path: 'src', fail-on-findings: 'false' }
- uses: github/codeql-action/upload-sarif@v3
  with: { sarif_file: ${{ steps.impeccable.outputs.sarif-file }} }
```

Outputs `sarif-file` and `finding-count`; `fail-on-findings: 'true'` blocks the job on any finding. Includes an example `workflow_dispatch` workflow and a README "Continuous integration" section.

---

> [!NOTE]
> **Low Risk**
> Additive CI and output paths with no auth or data-handling changes; optional job failure is opt-in via `fail-on-findings`.
> 
> **Overview**
> Adds **CI integration** so design anti-pattern scans can feed **GitHub code scanning** via SARIF 2.1.0.
> 
> The CLI gains **`detect --sarif`**, which writes SARIF to stdout through new **`toSarif()`** (`cli/engine/output/sarif.mjs`): severities map to SARIF levels, paths are relativized, rules dedupe into `tool.driver.rules`, and **`category` / `engine`** land in rule properties. **SARIF mode always exits 0** (even with findings) so a follow-up upload step still runs; **`--json`** behavior is unchanged.
> 
> A **composite reusable Action** (`action.yml`) runs `npx impeccable@latest detect --sarif`, writes `impeccable.sarif`, exposes **`sarif-file`** and **`finding-count`**, and optional **`fail-on-findings`**. README documents the workflow; an example **`workflow_dispatch`** job uploads SARIF with `github/codeql-action/upload-sarif@v3`.
> 
> Supporting changes: findings gain **`category`** and per-engine **`engine`** tags (regex, static-html, browser); human-readable CLI output can show them; shared numeric thresholds move to **`shared/thresholds.mjs`**; public **`index.d.ts`** and package **`types`** export; **`toSarif`** is re-exported from the API facade. New tests cover SARIF conversion, CLI `--sarif` e2e, provenance, thresholds, and types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f26afb0de9f61aa6f03aa4dc9c23f49583221f13. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pbakaus (changes requested):
Thanks for putting this together. I like the direction: SARIF output plus a reusable Action would make the detector much more useful in CI.

I don’t think this is merge-ready yet, though.

Main blockers:

1. The Action runs `npx --yes impeccable@latest`, not the CLI from the same ref as the Action. I verified that current `impeccable@latest` does not behave like this branch: `npx --yes impeccable@latest detect --sarif demos` printed normal human output and exited non-zero instead of writing SARIF. So the advertised Action path currently breaks before it can upload SARIF or set outputs.

2. The branch is stale/conflicted against current `main`.

3. The provenance issue Cursor flagged is real. `detectHtml()` calls the regex text-content analyzers, then sets every finding to `engine = 'static-html'`, so HTML text-content findings like `em-dash-overuse` and `marketing-buzzword` are mislabeled.

The core SARIF implementation itself looks promising: the focused SARIF/provenance/threshold/type tests pass, and the branch’s local `detect --sarif demos` emits parseable SARIF.

PR #220, opened 2026-06-07
fix(live-server): stop cross-origin token leak + add DNS-rebinding guard

mtthwgrvn:
## Problem

The live variant server (`skill/scripts/live-server.mjs`) serves `/live.js` with `Access-Control-Allow-Origin: *` and no auth. That response body embeds the per-session token:

```js
window.__IMPECCABLE_TOKEN__ = '<uuid>';
```

Because the server binds `127.0.0.1` and advertises a wildcard CORS origin, **any website open in the same browser** can:

1. `fetch('http://127.0.0.1:<port>/live.js')` — the wildcard ACAO permits reading the cross-origin response body.
2. Parse the session token out of it.
3. POST to the token-gated write/apply endpoints, writing attacker-controlled content into files under the project `cwd`.

The port is discoverable by scanning from 8400. Written source files then execute on the developer's next build / dev-server reload → code execution on the dev host. Requires only that live mode is running and the developer visits a hostile page.

## Why the wildcard CORS isn't needed

The legitimate consumer loads `/live.js` via a `<script>` tag. Script-tag execution exposes the token to the requesting page **without** CORS read permission, so the real flow never depended on `Access-Control-Allow-Origin: *`. Removing it breaks nothing legitimate and closes the cross-origin `fetch()` read.

## Fix

- Remove `Access-Control-Allow-Origin: *` from the request handler.
- Add a loopback `Host`-header allowlist (`127.0.0.1` / `localhost` / `[::1]` : port). This blocks DNS-rebinding, where a rebound hostname resolving to `127.0.0.1` would otherwise make a hostile page same-origin and let it read the token even without CORS.
- Add regression tests covering the Host guard (403 + no token in body) and the absence of wildcard CORS.

## Testing

`node --test tests/live-server.test.mjs` — the two new tests pass. No change to the pass/fail set of the existing suite vs. baseline.

---

> [!NOTE]
> **High Risk**
> Security fix on authentication boundaries for a local server that embeds session tokens and exposes file-writing APIs; behavior change could break non-standard clients that relied on wildcard CORS or non-loopback Host headers.
> 
> **Overview**
> Closes a **local live-server exposure** where any website could read the session token from `/live.js` and drive token-gated write endpoints.
> 
> **`live-server.mjs`** drops **`Access-Control-Allow-Origin: *`** so cross-origin `fetch()` cannot read `/live.js` (legitimate use still loads it via `<script>`). Every request is gated by **`isAllowedLoopbackHost`**, which returns **403** unless `Host` is `127.0.0.1`, `localhost`, or `[::1]` with the bound port—blocking DNS-rebinding attacks that would otherwise reach token-bearing routes.
> 
> **`live-server.test.mjs`** adds **`rawRequest`** (custom `Host` via `node:http`) plus regression tests for the Host guard and absence of wildcard CORS.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c15b1eb67b7ddb11f1789206719bd708c65ce83a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skill/scripts/live-server.mjs
+++ b/skill/scripts/live-server.mjs
@@ -60,6 +60,18 @@ const SSE_HEARTBEAT_INTERVAL = 30_000;  // keepalive ping every 30s
 // Port detection
 // ---------------------------------------------------------------------------
 
+// Accept only loopback Host headers matching this server's port. Guards against
+// DNS-rebinding: the server binds 127.0.0.1, so a real request's Host is always
+// 127.0.0.1:<port> or localhost:<port>. A port-less Host (e.g. default :80) can
+// never legitimately reach us, so it is rejected too.
+function isAllowedLoopbackHost(hostHeader, port) {
+  if (!hostHeader || typeof hostHeader !== 'string') return false;
+  const host = hostHeader.toLowerCase();
+  return host === `127.0.0.1:${port}` ||
+         host === `localhost:${port}` ||
+         host === `[::1]:${port}`;
+}
+
 async function findOpenPort(start = 8400) {
   return new Promise((resolve) => {
     const srv = net.createServer();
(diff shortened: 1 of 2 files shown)


pbakaus:
Thanks for digging into this. I agree the underlying issue is real, and the Host-header allowlist is a good direction for the DNS-rebinding side.

I don’t think this can merge as-is, though. Removing `Access-Control-Allow-Origin` globally breaks the normal browser flow. `/live.js` still loads via `<script>`, but the loaded `live-browser.js` then calls the helper server from the dev app origin, which is a different origin because the port differs. That includes `fetch()` calls to `/manual-edit-stash`, `/status`, design-system endpoints, POSTs to `/events`, and the `EventSource` connection to `/events`.

I verified this manually. The new Node tests pass because Node fetch does not enforce browser CORS.

Suggested path:
- Keep the loopback Host guard.
- Replace wildcard CORS with a narrow Origin allowlist for loopback origins only, e.g. `http://localhost:<dev-port>`, `http://127.0.0.1:<dev-port>`, and `[::1]` variants as appropriate.
- Echo the allowed Origin instead of `*`, set `Vary: Origin`, and reject/omit CORS for non-loopback origins.
- Add a regression that covers a browser-like cross-origin localhost request, not just Node fetch.

So: the security fix is worth pursuing, but this version regresses live mode and needs another pass before merge.

pbakaus:
As there hasn't been any activity for 5 days, I'll close this PR for now. If you'd like to finish it based on the guidance I gave, feel free to reopen.

PR #230, opened 2026-06-08
fix(live-inject): preserve the character after an insertAfter anchor (#227)

### LINKED ISSUE #227: live-inject.mjs insertTag() drops a character after the anchor (off-by-one in insertAfter)
The `insertAfter` branch of `insertTag()` in `skills/impeccable/scripts/live-inject.mjs` (v3.5.0 bundle) has an off-by-one that corrupts the entry file during live-mode injection.

After `const after = idx + config.insertAfter.length;`, the code builds `prefix` — appending a `\n` when the anchor isn't already followed by one — then returns `prefix + block + content.slice(prefix.length)`. In the no-trailing-newline case `prefix.length === after + 1`, so `content.slice(prefix.length)` skips the first real character after the anchor — e.g. `<head>X…` loses the `X`. (The `insertBefore` branch and the anchor-already-followed-by-`\n` case are unaffected.)

**Repro**

```js
insertTag('<head>X</head>', { insertAfter: '<head>', /* …commentStyle etc… */ }, 5173, 'index.html');
// BUG: the 'X' immediately after <head> is dropped.
```

**Fix** — slice the remainder from the original `after` index, not `prefix.length`:

```js
const hadNewline = content[after] === '\n';
const prefix = hadNewline ? content.slice(0, after + 1) : content.slice(0, after) + '\n';
const rest = content.slice(hadNewline ? after + 1 : after);
return prefix + block + rest;
```

Happy to open a PR. Found via a code review on a downstream project using the v3.5.0 bundle.

Osamaali313:
## Summary

Fixes #227.

`insertTag()` in `skill/scripts/live-inject.mjs` has an off-by-one in its
`insertAfter` branch. After computing `after = idx + config.insertAfter.length`,
it builds `prefix` (appending a `\n` when the anchor isn't already followed by
one) and returns `prefix + block + content.slice(prefix.length)`.

In the no-trailing-newline case `prefix.length === after + 1`, so
`content.slice(prefix.length)` skips the first real character after the anchor —
e.g. injecting after `<head>` into `<head>X…` drops the `X`, corrupting the
entry file during live-mode injection.

## Fix

Slice the remainder from the original anchor offset instead of `prefix.length`:

```js
const hadNewline = content[after] === '\n';
const prefix = hadNewline ? content.slice(0, after + 1) : content.slice(0, after) + '\n';
const rest = content.slice(hadNewline ? after + 1 : after);
return prefix + block + rest;
```

The `insertBefore` branch and the anchor-already-followed-by-`\n` case are
unchanged.

## Tests

Adds two cases to `tests/live-inject.test.mjs` exercising `insertTag` directly:
- the no-trailing-newline case (`<head>X</head>`) — fails before the fix, passes
  after;
- the already-followed-by-newline case — guards against regressions in the
  unaffected branch.

`node --test tests/live-inject.test.mjs` → 12/12 passing.

## Generated bundles

Regenerated the tracked per-agent copies (`.agents`, `.claude`, `.cursor`,
`.gemini`, `.github/skills`, `.kiro`, `.opencode`, `.pi`, `.qoder`, `.rovodev`,
`.trae`, `.trae-cn`, `plugin/`) so the fix ships everywhere and the
"Verify generated tracked outputs" CI check stays green. Each copy is
byte-identical to the source.

## Credit

Root-cause analysis (and the suggested fix) came from the #227 reporter.

---

> [!NOTE]
> **Medium Risk**
> Changes deterministic edits to user HTML entry files during live mode; wrong slicing could corrupt pages, though scope is narrow and covered by new tests.
> 
> **Overview**
> Fixes **#227** in `live-inject.mjs`: when using `insertAfter` and the anchor is not followed by a newline (e.g. `<head>X</head>`), the old logic sliced the remainder at `prefix.length` and **dropped the first character** after the anchor. The remainder is now taken from the anchor offset (`after + existingNewline.length`).
> 
> Also adds **line-ending awareness**: detect CRLF/LF/CR in the file, normalize the injected block to match, and broaden `removeTag` patterns so inject/remove round-trips on Windows-style files.
> 
> The same script update is copied to all tracked per-agent `live-inject.mjs` bundles. **Tests** cover the compact-anchor case and CRLF inject/remove.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ebf4355d355a868e464f48ef1052e4da9f771d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/skills/impeccable/scripts/live-inject.mjs
+++ b/.agents/skills/impeccable/scripts/live-inject.mjs
@@ -368,8 +368,26 @@ function buildTagBlock(syntax, port, filePath) {
   );
 }
 
+function detectLineEnding(content) {
+  if (content.includes('\r\n')) return '\r\n';
+  if (content.includes('\r')) return '\r';
+  return '\n';
+}
+
+function normalizeLineEndings(content, lineEnding) {
+  return lineEnding === '\n' ? content : content.replace(/\n/g, lineEnding);
+}
+
+function readLineEndingAt(content, index) {
+  if (content[index] === '\r' && content[index + 1] === '\n') return '\r\n';
+  if (content[index] === '\n') return '\n';
+  if (content[index] === '\r') return '\r';
+  return '';
+}
+
 function insertTag(content, config, port, filePath) {
-  const block = buildTagBlock(config.commentSyntax, port, filePath);
+  const lineEnding = detectLineEnding(content);
+  const block = normalizeLineEndings(buildTagBlock(config.commentSyntax, port, filePath), lineEnding);
   // insertBefore: match the LAST occurrence. Anchors like `</body>` naturally
   // belong at the end, and the same literal can appear earlier in code blocks
   // within rendered documentation pages.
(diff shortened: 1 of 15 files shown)


pbakaus:
Thanks for jumping on this and for the clean regression coverage around #227. I took the liberty of pushing a small follow-up commit to improve the newline handling a bit further: the insertAfter path now preserves CRLF/CR/LF boundaries, the injected block follows the file's newline style, and the tests exercise the public live-inject CLI path including a CRLF round trip. Appreciate you catching the original off-by-one!

PR #271, opened 2026-06-19
Fix live toast stale callback race

### LINKED ISSUE #268: [Bug] Live mode: unguarded requestAnimationFrame in showToast() throws "Cannot read properties of null (reading 'style')"
## What happened?

During a live session, the bundled live script throws an uncaught error that recurs while the picker is active:

    live.js:8053 Uncaught TypeError: Cannot read properties of null (reading 'style')

It fires on pages where part of the UI re-renders frequently (for example a subtree updated on a polling/refetch interval) while a live session runs, and it disrupts the picker on that element. Root cause: an unguarded `requestAnimationFrame` in `showToast()` that reads `toastEl.style` after `toastEl` has been set to `null`. Full trace and a one-line fix are in Additional context.

## Steps to reproduce

1. Run a live session on a Vite + React page.
2. Pick an element inside a subtree that re-renders on a recurring interval (e.g. a React Query `refetchInterval`).
3. Watch the console throw `Cannot read properties of null (reading 'style')` from the bundled `live.js`, recurring with the re-render cadence.

## Expected behavior

The toast helper should not crash. The enter-animation `requestAnimationFrame` in `showToast()` should null-guard `toastEl` the same way its sibling `setTimeout` already does, so a concurrent `dismissToast()` cannot cause an uncaught `TypeError`.

## Provider & environment

- **Provider** (Cursor / Claude Code / Gemini CLI / Codex / Copilot / Kiro / OpenCode): Claude Code
- **Provider version**: (optional)
- **OS**: Windows (the bug is browser-side and OS-independent)
- **impeccable version**: 3.5.0

## Additional context

Trace (v3.5.0): `live.js:8053` maps to `showToast()` in `live-browser.js` (around source line 7776). The toast enter-animation runs in a `requestAnimationFrame` that reads the module-scoped `toastEl` with no null check:

    // live-browser.js, showToast() (~line 7776)
    requestAnimationFrame(() => {
      toastEl.style.opacity = '1';                              // throws here
      toastEl.style.transform = 'translateX(-50%) translateY(0)';
    });

The sibling auto-dismiss `setTimeout` right below already guards the same variable (`if (toastEl) { ... }`). `toastEl` is set to `null` without reassignment by `dismissToast()`, reached from `showBar('generating')` via `if (recoveryWaitingForAnchor) dismissToast();` (~line 1101). When `dismissToast()` runs in the ~1 frame between the rAF being scheduled and firing, `toastEl` is `null` and the rAF throws.

Why frequent DOM churn triggers it: `enterRecoveryWaitingForAnchor()` installs a `MutationObserver` on `document.body` with `{ childList: true, subtree: true }`. On a page whose UI re-renders on a timer, those mutations keep resolving the anchor and driving `showBar('generating')` -> `dismissToast()` while a toast enter-rAF is still pending, so the race is hit repeatedly rather than rarely.

Suggested fix: guard the rAF like the sibling `setTimeout` already does:

    requestAnimationFrame(() => {
      if (!toastEl) return;     // add this
      toastEl.style.opacity = '1';
      toastEl.style.transform = 'translateX(-50%) translateY(0)';
    });

One line, matches the existing guard pattern, removes the crash regardless of churn.

jjoanna2-debug:
## Summary
- bind live-mode toast rAF/timeout callbacks to the toast element that scheduled them
- prevent `dismissToast()` or a newer toast from leaving stale callbacks that touch `toastEl`
- add a regression guard for the #268 timing race and stale timer ownership

## Why
Fixes #268.

`showToast()` schedules an enter `requestAnimationFrame`, but `dismissToast()` can remove the current toast and set `toastEl = null` before that frame runs. The old callback then read `toastEl.style` and could throw. A plain null guard would close that crash, but it still leaves older timeout callbacks able to animate or remove a newer toast.

This patch captures the created element as `currentToast` and makes each delayed callback no-op unless `toastEl === currentToast`.

## Verification
- `node --test tests/live-browser-regression.test.mjs`
- `node --test tests/live-browser-regression.test.mjs tests/live-browser-source.test.mjs tests/live-browser-dom.test.mjs tests/live-browser-script-parts.test.mjs tests/live-browser-session.test.mjs`
- `git diff --check`
- `npx --yes bun run test:live` still fails on `tests/live-copy-edit-agent.test.mjs` (`flags invalid JSX syntax in post-apply checks`); confirmed the same failure on clean `origin/main`.
- `npx --yes bun run test:live-e2e` is blocked in this checkout because `@anthropic-ai/sdk` is not installed.

---

> [!NOTE]
> **Low Risk**
> Scoped UI toast lifecycle fix with regression tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes a timing race in live-mode **`showToast()`** where enter **`requestAnimationFrame`** and auto-dismiss **`setTimeout`** callbacks could run after **`dismissToast()`** or a newer toast replaced the global **`toastEl`**, causing crashes or animating/removing the wrong toast.
> 
> Each toast now binds delayed work to a local **`currentToast`**; enter, fade-out, and removal callbacks no-op unless **`toastEl === currentToast`**. Regression tests in **`live-browser-regression.test.mjs`** lock in that ownership behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31a5a56007671033c342233c866f96cffa227352. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skill/scripts/live-browser.js
+++ b/skill/scripts/live-browser.js
@@ -7916,7 +7916,7 @@ void main() {
     const barTopFromBottom = barRect && barRect.height > 0
       ? Math.max(16, window.innerHeight - barRect.top + 12)
       : 16;
-    toastEl = el('div', {
+    const currentToast = el('div', {
       position: 'fixed', bottom: barTopFromBottom + 'px', left: '50%',
       transform: 'translateX(-50%) translateY(8px)',
       background: C.ink, color: C.white,
@@ -7926,19 +7926,24 @@ void main() {
       transition: 'opacity 0.25s ' + EASE + ', transform 0.25s ' + EASE,
       pointerEvents: 'none', maxWidth: '420px', textAlign: 'center',
     });
-    toastEl.id = PREFIX + '-toast';
-    toastEl.textContent = message;
-    uiAppend(toastEl);
+    toastEl = currentToast;
+    currentToast.id = PREFIX + '-toast';
+    currentToast.textContent = message;
+    uiAppend(currentToast);
     requestAnimationFrame(() => {
-      toastEl.style.opacity = '1';
-      toastEl.style.transform = 'translateX(-50%) translateY(0)';
+      if (toastEl !== currentToast) return;
+      currentToast.style.opacity = '1';
+      currentToast.style.transform = 'translateX(-50%) translateY(0)';
     });
     setTimeout(() => {
-      if (toastEl) {
-        toastEl.style.opacity = '0';
-        toastEl.style.transform = 'translateX(-50%) translateY(8px)';
-        setTimeout(() => { if (toastEl) { toastEl.remove(); toastEl = null; } }, 250);
-      }
+      if (toastEl !== currentToast) return;
+      currentToast.style.opacity = '0';
+      currentToast.style.transform = 'translateX(-50%) translateY(8px)';
+      setTimeout(() => {
+        if (toastEl !== currentToast) return;
+        currentToast.remove();
+        toastEl = null;
+      }, 250);
     }, duration);
   }
 
--- a/tests/live-browser-regression.test.mjs
+++ b/tests/live-browser-regression.test.mjs
@@ -490,6 +490,34 @@ describe('live-browser.js regression guards', () => {
     );
   });
 
+  it('toast enter and dismiss timers only touch the current toast element', () => {
+    assert.match(
+      SOURCE,
+      /function showToast\(message, duration\)[\s\S]{0,1200}?const currentToast = el\('div'/,
+      'showToast must capture the created toast in a local so delayed callbacks cannot dereference or remove a stale global toastEl',
+    );
+    assert.match(
+      SOURCE,
+      /requestAnimationFrame\(\(\) => \{[\s\S]{0,80}?if \(toastEl !== currentToast\) return;[\s\S]{0,120}?currentToast\.style\.opacity = '1';/,
+      'toast enter rAF must no-op when dismissToast or a newer toast replaced toastEl before the frame fires',
+    );
+    assert.match(
+      SOURCE,
+      /setTimeout\(\(\) => \{[\s\S]{0,80}?if \(toastEl !== currentToast\) return;[\s\S]{0,120}?currentToast\.style\.opacity = '0';/,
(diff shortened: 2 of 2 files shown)


abdulwahabone:
I am not able to repro this bug, can you share a minimal repo for us to test and repro it.

jjoanna2-debug:
Sure. I added a minimal dependency-free repro repo here:

https://github.com/jjoanna2-debug/impeccable-toast-race-repro

Steps:

```sh
git clone https://github.com/jjoanna2-debug/impeccable-toast-race-repro.git
cd impeccable-toast-race-repro
python3 -m http.server 4173
```

Then open `http://localhost:4173` and click **Crash old implementation**.

What it does:

1. Calls `showToastOld()`.
2. Immediately calls `dismissToastOld()` before the next animation frame.
3. The pending `requestAnimationFrame` callback then runs after `oldToastEl` has been set to `null`.
4. The old callback reads `oldToastEl.style` and throws `Cannot read properties of null (reading 'style')`.

The same page has **Run fixed implementation**, which uses the `currentToast` identity guard from this PR. That path no-ops the stale callback instead of touching a nulled or superseded toast.

I kept it static and framework-free so the repro isolates the toast lifecycle race itself rather than depending on a particular React/Vite page shape.

jjoanna2-debug:
Quick status refresh: this is still current and green.

I re-ran the focused regression guard locally on Node v26.4.0:

```sh
node --test tests/live-browser-regression.test.mjs
```

Result: 28 pass, 0 fail.

The minimal repro is still here:
https://github.com/jjoanna2-debug/impeccable-toast-race-repro

It isolates the failing sequence: `showToastOld()` schedules the enter `requestAnimationFrame`, then `dismissToastOld()` clears `toastEl` before that callback runs. The PR fix captures `currentToast` and makes stale callbacks no-op unless the global toast reference still points at the same element.

Current PR checks are green, including `changes`, `test`, live-e2e smoke jobs, and Cursor Bugbot. Scope is limited to the live toast lifecycle; no API, auth, data, or persistence surface is touched.

pbakaus:
Seems sensible - thank you for your contribution!

PR #277, opened 2026-06-20
Don't force init on scoped commands when PRODUCT.md is missing

Rob-Makappen:
## Problem

Setup step 1 in `SKILL.md` says:

> **If it reports `NO_PRODUCT_MD`, stop and follow `reference/init.md` before doing anything else.**

In any project without a `PRODUCT.md`, this turns *every* scoped request into a full from-scratch `init` detour. Ask for `/impeccable polish index.html` and the skill instead stops and starts writing `PRODUCT.md` from the beginning. `context.mjs` reinforces it by printing "Stop the current task, load reference/init.md...".

Two failure modes follow:
1. The agent runs the whole unwanted init onboarding for a one-line request.
2. Faced with the heavy gate, the agent abandons the command entirely and does an ad-hoc pass without ever loading the command reference, so the skill effectively doesn't fire.

This bites hardest on real, existing codebases (the common case for `polish` / `critique` / `audit` / `layout` / etc.), where there's plenty of context in the code itself but no `PRODUCT.md` was ever authored.

## Fix

Make the gate **command-aware** rather than blanket:

- **From-scratch build flows** (`init`, `craft`, `shape`) still route into `reference/init.md` when `PRODUCT.md` is missing. That's where captured product context genuinely matters before building.
- **Every other (scoped) command against existing code** no longer diverts. The code is the context: proceed with the requested command, infer the register from the surface in focus (step 4), and offer `/impeccable init` once as a *suggestion*, never a blocker.

## Changes

- `skill/SKILL.src.md` — rewrite the step 1 `NO_PRODUCT_MD` rule; reconcile the no-argument routing rule so it *leads the menu* with `init` instead of silently jumping into it.
- `skill/scripts/context.mjs` — soften the `NO_PRODUCT_MD` message to defer to the step 1 rule. Still prefixed `NO_PRODUCT_MD:` and still references `reference/init.md`.
- `tests/skill-behavior/scenarios.test.mjs` — add **scenario 10**: a scoped command (`/impeccable polish`) with no `PRODUCT.md` loads `polish.md` and proceeds, and must **not** divert into `init.md`. Scenario 1 (a from-scratch `craft` still correctly diverts to init) stays green and pins that path.
- Regenerated provider + plugin copies via `bun run build:skills`.

## Testing

- `node --test tests/context.test.mjs` — 30 pass
- `bun test tests/build.test.js` (generated-copies sync check) — 10 pass
- `tests/skill-behavior/scenarios.test.mjs` parses and skips cleanly without API keys (new scenario runs in the keyed CI sweep)
- Build passes both prose validators

---

> [!NOTE]
> **Low Risk**
> Changes are skill prose, boot messaging, and LLM behavior tests; they refine agent routing without touching runtime product code or security-sensitive paths.
> 
> **Overview**
> **Missing `PRODUCT.md` no longer blocks scoped work** like `polish` / `audit` on existing code. Setup step 1 and `context.mjs` now branch on intent: **from-scratch** flows (`init`, `teach`, `craft`, `shape`, or clear build/shape wording) still load `reference/init.md` first; **other commands** proceed using the codebase as context, infer register from the target surface, and only **suggest** `/impeccable init`.
> 
> **Menu routing** when `NO_PRODUCT_MD` is reported now **recommends** init at the top instead of silently entering init. **`shape`** is aligned with **`craft`** in the init-blocker note.
> 
> **Behavior tests** add scenarios 10–13 (scoped polish without init; shape, natural-language build, and `teach` still divert to init) and document them in the skill-behavior README.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb054ced06b1a214b53cb28f10ca60d6668dbdf9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pbakaus:
I was very much on the fence about this, but I'm coming around to the idea. A command like "bolder" doesn't strictly need to know the audience. That being said, could you please remove the generated files from the PR and just have the source files in the source folder in this PR (repo convention), plus, please clear any of the bugbot review items (if they're false positives, say so and dismiss). Happy to take a look afterwards, thanks!

Rob-Makappen:
Thanks Paul, that makes sense. Both done:

1. Generated files removed - the PR is now source-only (`skill/SKILL.src.md`, `skill/scripts/context.mjs`, and the test), and I rebased onto latest main so it's no longer conflicting. The provider and plugin copies will regenerate from your `build:skills` sync.
2. The three Greptile items were all genuine, not false positives, so I fixed rather than dismissed:
    - `context.mjs` file-level JSDoc still described the old empty-stdout signal, now refreshed to the `NO_PRODUCT_MD:` behaviour.
    - the `craft` init-then-resume footnote now covers `shape` too, since it's a from-scratch flow as well.
    - added scenario 11 (shape + no PRODUCT.md still diverts to init) alongside scenario 10 (scoped command proceeds).

`build:skills` is green (version and prose validators pass). Ready for another look, thanks.

pbakaus (approved):
sorry for the wait, I took the liberty to fix the remaining issues, thanks for your contribution!

PR #282, opened 2026-06-20
Fix monorepo target-selection edge cases (#213 review follow-up)

pbakaus:
Follow-up to the merged #213 (monorepo context support), addressing the unresolved review comments.

## Cursor Bugbot (Medium) — fixed

1. **Excluded packages still listed** (`context.mjs` `discoverTargetCandidates`). The candidate list applied positive workspace globs but never the negated ones, so an excluded package (`"!packages/internal"`) appeared as a selectable target — yet picking it resolves back to the repo root (`resolveWorkspaceProjectRoot` already honors the exclusion). Now filtered with the same `isExcludedByWorkspacePattern` check the resolver uses.

2. **Empty app list blocks root** (`context.mjs` `resolveTargetSelection`). It returned `TARGET_SELECTION_REQUIRED` whenever `projectRoot === repoRoot`, even with **zero** discoverable child apps (e.g. `workspaces: ["."]`, or a marker file with no `apps/`/`packages/` children). That left an unanswerable selection prompt. It now returns `null` (treat the repo root as the project) when there are no candidates.

## Greptile (P2) — documented

3. `contextSourceStatus`: added a comment spelling out the four statuses and the deliberate dual meaning of `'fallback'` (a context file in a project subdirectory via `FALLBACK_DIRS`, and one outside both roots via `IMPECCABLE_CONTEXT_DIR`).
4. `findMonorepoRoot`: documented why `isMonorepoRoot` is intentionally checked before `hasGitBoundary` (a nested `.git` is an independent project boundary).

## Tests

Two regression tests in `tests/context.test.mjs`: negated packages are excluded from the candidate list, and a root-only workspace does not block on selection. Full default suite green (198 pass, 0 fail); build passes prose + count validators. Source-first.

---

> [!NOTE]
> **Low Risk**
> Scoped monorepo discovery and selection logic with regression tests; no auth, security, or broad API surface changes.
> 
> **Overview**
> Fixes two monorepo **target selection** bugs in `context.mjs` and adds short comments for existing resolution behavior.
> 
> **Negated workspace packages** no longer appear in `discoverTargetCandidates`. Candidates are filtered with `isExcludedByWorkspacePattern`, matching `resolveWorkspaceProjectRoot`, so excluded paths like `!packages/internal` cannot be picked only to resolve silently to the repo root.
> 
> **Root-only monorepos** (`workspaces: ["."]` or markers with no `apps`/`packages` children) skip `TARGET_SELECTION_REQUIRED` when there are zero candidates; context loads from the repo root instead of an empty prompt.
> 
> Comments document `contextSourceStatus` statuses (including dual `'fallback'`) and why `findMonorepoRoot` checks `isMonorepoRoot` before `.git`. **Tests** cover negated-package exclusion and the no-child-apps path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4efed40b083e02491d4bf3469c22c61df1ef9ad5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skill/scripts/context.mjs
+++ b/skill/scripts/context.mjs
@@ -135,11 +135,17 @@ export function resolveTargetSelection(cwd = process.cwd(), options = {}) {
   ) {
     return null;
   }
+  const targetCandidates = discoverTargetCandidates(project.repoRoot);
+  // No discoverable child apps (e.g. `workspaces: ["."]`, a root-only workspace,
+  // or a marker file with no apps/packages children): there is nothing to choose,
+  // so treat the repo root as the active project rather than blocking on an empty
+  // selection prompt that the user cannot answer.
+  if (targetCandidates.length === 0) return null;
   return {
     targetPath: null,
     projectRoot: project.projectRoot,
     repoRoot: project.repoRoot,
-    targetCandidates: discoverTargetCandidates(project.repoRoot),
+    targetCandidates,
   };
 }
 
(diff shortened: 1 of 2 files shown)


abdulwahabone:
Thanks for fixing this edge case @pbakaus

PR #300, opened 2026-06-25
Add COPY.md support via document-copy command

harsh51191:
## Summary

- Adds **`document-copy`** — a new Impeccable build command that generates `COPY.md`, an experience language system (voice, journey playbooks, microcopy patterns, i18n rules) complementing the existing `PRODUCT.md` + `DESIGN.md` trio.
- Extends **`context.mjs`** and **`context-signals.mjs`** to load and surface `COPY.md` alongside product and design context.
- Ships **`scan-copy.mjs`** for locale/i18n evidence gathering (mechanics and debt signals — not voice), plus register references (`product-copy.md`, `brand-copy.md`) and supporting docs (`copy-journeys.md`, `copy-decision-trees.md`, `copy-ai-trust.md`).
- Wires **`init`**, **`clarify`**, and routing to recommend `document-copy` when code exists but COPY.md is missing.

**Impeccable trio:** `PRODUCT.md` = who/why · `DESIGN.md` = how it looks · **`COPY.md` = how it reads across journeys.**

All examples and reference content use generic SaaS patterns (task apps, billing, analytics, workspace admin). No customer-specific copy or fixtures are included.

## Test plan

- [ ] `node --test tests/context.test.mjs` — COPY.md resolution and CLI output
- [ ] `node --test tests/scan-copy.test.mjs` — locale scanner
- [ ] On any repo with `PRODUCT.md` and i18n files: run `/impeccable document-copy` and confirm interview + COPY.md generation flow
- [ ] Run `node skill/scripts/context.mjs` and verify `# COPY.md` block appears when file exists
- [ ] Run `node skill/scripts/scan-copy.mjs --target <app>` and confirm JSON evidence output (not voice rules)

---

> [!NOTE]
> **Low Risk**
> Changes are skill docs, context loading, and a read-only locale scanner with tests—no auth, payments, or production runtime paths beyond agent guidance.
> 
> **Overview**
> Adds **`document-copy`** as the 24th Impeccable command and introduces **`COPY.md`** as the experience-language counterpart to `PRODUCT.md` and `DESIGN.md` (voice, journeys, patterns—not a string dump).
> 
> **`context.mjs`** now resolves, loads, and prints `COPY.md` (including fallback dirs and monorepo selection); **`context-signals.mjs`** exposes `hasCopy` / `copyPath`. New **`scan-copy.mjs`** gathers locale/i18n evidence for the command flow, with **`tests/scan-copy.test.mjs`** in the core suite.
> 
[...]

PR #307, opened 2026-06-25
Allow context roots to be declared in .impeccable/config.json (decoupled from package managers)

CypherPoet:
> **Note:** description refreshed by the maintainer (via Claude Code) after the second review round, so it matches what the branch actually ships. The original said `contextRoots` and left the naming open; the code landed on `projectRoots`. No code was changed.

## Why

After reading #300, I initially assumed that I could drop a `DESIGN.md` or `PRODUCT.md` into any folder of my project and have Impeccable treat it as a "monorepo app". Digging in, that capability is tightly coupled to package-manager configuration: a repo is only recognized as a monorepo when `package.json` `workspaces`, `pnpm-workspace.yaml`, or `lerna.json` (or a marker file beside `apps/`/`packages/`) declares the workspaces.

This PR closes that gap.

## What

Adds a `projectRoots` glob list to `.impeccable/config.json` (shared) and `.impeccable/config.local.json` (per-developer):

```json
{ "projectRoots": ["docs/design/throwback-jerseys/*"] }
```

With it, a repo containing **no** package-manager files is still recognized as a monorepo, and each declared child resolves its own `PRODUCT.md`/`DESIGN.md` with per-file fallback to the root, identical to how package-manager workspaces already behave. Same glob syntax as `workspaces`, including `!` negation, and it **composes with** package-manager workspaces rather than replacing them.

## How

Project boundaries now come from two ordered groups instead of one flat list:

```js
function readProjectPatternGroups(repoRoot) {
  return [
    readImpeccableProjectRoots(repoRoot),        // new
    [
      ...readPackageWorkspaces(repoRoot),
      ...readPnpmWorkspaces(repoRoot),
      ...readLernaWorkspaces(repoRoot),
    ].filter(Boolean),
  ];
}
```

**Precedence.** A path matched by any `projectRoots` pattern, positive or negated, is governed by that group alone. Package-manager patterns apply only to paths `projectRoots` does not match, and each group's `!` negations are scoped to its own source. So an Impeccable negation can hide a package workspace from Impeccable, while a package-level negation never hides a folder that `projectRoots` declares.

`isSelectableCandidate()` keeps the app picker consistent with that rule: a folder is only offered when picking it would resolve back to itself, so a negated path, or one whose boundary sits at a shallower `projectRoots` match, is not listed as a target that would silently resolve elsewhere.

`readImpeccableProjectRoots()` reads `projectRoots` from `.impeccable/config.json` then `config.local.json` (per-dev extends shared). An `.impeccable` config without a `projectRoots` key changes nothing.

Detection (`isMonorepoRoot`), project resolution (`resolveWorkspaceProjectRoot`), and the app picker (`discoverTargetCandidates`) all funnel through the new aggregator, so no other call site changed.

## Tests

10 new cases in `tests/context.test.mjs`. Five for the feature itself:

- detection with **no** package-manager files present
- `--target` resolution and run-from-inside-the-folder resolution
- `config.local.json` extending shared `projectRoots`
- no regression for an `.impeccable` config without `projectRoots` (stays single-project)
- the root app-picker listing config-declared children

Five for composition with package-manager workspaces:

- an Impeccable negation excluding a package-manager workspace
- a package-manager negation staying scoped to its own source
- a broad Impeccable pattern taking the boundary over a deeper package workspace
- fall-through to package workspaces for paths Impeccable does not match
- other workspaces resolving normally when the Impeccable config only negates

## Docs

`site/content/reference/config.md` gains a "Project Roots" section covering path relativity, glob and negation syntax, how shared and local config combine, and the precedence rule above. `site/content/reference/context.md` notes the second source of project boundaries.

## Notes

Source-only diff: four files, no generated provider output staged, per the source-first policy in `CLAUDE.md`.

### DIFF
--- a/site/content/reference/config.md
+++ b/site/content/reference/config.md
@@ -12,7 +12,8 @@ Use config for:
 
 - detector ignores shared by `npx impeccable detect` and the design hook;
 - private local ignores that should not be committed;
-- hook lifecycle settings such as enabled, quiet mode, and audit logging.
+- hook lifecycle settings such as enabled, quiet mode, and audit logging;
+- project roots, for repos where design boundaries are not declared by a package manager.
 
 Use `PRODUCT.md` and `DESIGN.md` for product and design intent. See [Design Context](/docs/context).
 
@@ -51,6 +52,24 @@ npx impeccable ignores add-file "src/private-experiment/**" --local
 ```
 
 Local settings go into `.impeccable/config.local.json`, which Impeccable keeps out of git.
+
+## Project Roots
+
+Impeccable normally finds nested projects through package-manager workspace declarations: `package.json` workspaces, `pnpm-workspace.yaml`, or `lerna.json`. When those files do not exist, or when design boundaries do not line up with packages, declare the roots directly:
+
+```json
+{
+  "projectRoots": ["docs/design/skins/*"]
+}
+```
+
+Each matched folder becomes its own project: it can carry its own `PRODUCT.md` and `DESIGN.md`, it appears in the app picker, and it falls back to the repo root per file for any context it does not define. See [Design Context](/docs/context).
+
+How the patterns behave:
+
+- Patterns are relative to the repo root and use the same glob syntax as `package.json` workspaces, including `*`, `**`, and `!` negation.
+- `projectRoots` in `config.local.json` extends the shared list, so one developer can add private roots without committing them.
+- A path matched by any `projectRoots` pattern, positive or negated, is governed by this config alone. Package-manager workspaces apply only to paths these patterns do not match, and each source's `!` negations apply only to its own patterns. So `"!apps/internal"` here hides a package workspace from Impeccable, while a package-level negation never hides a folder that `projectRoots` declares.
 
 ## Value ignores
 
--- a/site/content/reference/context.md
+++ b/site/content/reference/context.md
@@ -107,6 +107,7 @@ Treat context files like any other design artifact: review them in code review w
   <div>
     <p>For normal projects, put <code>PRODUCT.md</code> and <code>DESIGN.md</code> in the project root.</p>
     <p>Skill commands look in the root first. If root context is missing, they also check <code>.agents/context/</code> and <code>docs/</code>.</p>
+    <p>In a monorepo, each workspace child resolves its own <code>PRODUCT.md</code> and <code>DESIGN.md</code> first, then falls back to the repo root per file. Project boundaries come from package-manager workspace declarations, or from <code>projectRoots</code> globs in <code>.impeccable/config.json</code> when no package manager declares them. See <a href="/docs/config">Config and ignores</a>.</p>
     <p>The detector's design-system rules use the same root-first behavior for <code>DESIGN.md</code>. For generated design metadata, the primary path is <code>.impeccable/design.json</code>. Legacy <code>DESIGN.json</code> files are still accepted as fallbacks, but new projects should use <code>.impeccable/design.json</code>.</p>
   </div>
 </details>
--- a/skill/scripts/context.mjs
+++ b/skill/scripts/context.mjs
@@ -13,7 +13,9 @@
  *      canonical context files in an ordinary repo (issue #376).
  *   2. Active project .agents/context/ then docs/
  *   3. Repo root context, using the same order, as a per-file fallback
- *      whenever the active project is nested below it
+ *      whenever the active project is nested below it (a repo counts as a
+ *      monorepo when a package manager declares workspaces, or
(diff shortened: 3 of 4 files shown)


pbakaus (changes requested):
Thanks for this! I like the overall direction.

Before merging, I’d like to tighten two parts:

1. Define how Impeccable patterns compose with package-manager patterns. Right now all patterns are flattened together. Positive patterns use first-match behavior, while negations apply globally. That creates ambiguous cases when the two sources overlap: for example, an Impeccable negation can exclude a package-manager workspace, while a broad Impeccable pattern can change the boundary of a more specific package workspace. This behavior may be intentional, but because composition and `!` negation are advertised features, it should be explicitly defined and covered with tests. My inclination is that explicit Impeccable configuration should govern paths it matches, while package-manager discovery fills in everything else.
2. Document the new public configuration. Please add it to the context/config docs, including path relativity, glob and negation behavior, how shared and local config combine, and how it interacts with package workspaces.

I’m also leaning toward projectRoots rather than `contextRoots`. These patterns establish project boundaries and app-picker targets; “context roots” sounds more like directories where PRODUCT.md or DESIGN.md live. Internally, something like `readProjectPatterns()` would make the abstraction clearer, with package workspaces as one discovery source.

I’m not asking for a large refactor, the existing pipeline is the right architecture. I’d keep the implementation small, make the precedence explicit, add the missing interaction tests, and document the feature.

Otherwise, the code looks clean and focused. Thanks again.

CypherPoet:
@pbakaus Thanks for the feedback! All three points are now addressed.

### Naming

Adopted `projectRoots`, and the internal aggregator is now `readProjectPatternGroups()` / `readProjectPatterns()`, with package workspaces as one discovery source.


### Pattern Composition

Pattern composition is now explicit and matches your inclination: a path matched by any Impeccable pattern -- positive or negated -- is governed by the Impeccable group alone. Package-manager patterns only apply to paths the Impeccable group does not match, and `!` negations are scoped to their own source. 

One deliberate consequence in the app picker: a package workspace subsumed by a broader `projectRoots` boundary is no longer listed as a candidate, since choosing it would silently resolve to that boundary. 


### Testing 

Five new tests cover the interaction cases: impeccable negation excluding a package workspace, package negation not touching an impeccable root, broad-vs-specific boundary governance, fall-through for unmatched paths, and a negation-only config.


### Docs

Added a "Project Roots" section to the config reference (path relativity, glob/negation syntax, shared + local merge, precedence vs package workspaces) and a short monorepo note in the context reference's lookup details.


### Scope 

Implementation stayed small: same pipeline, with the flat pattern list split into two source groups evaluated in order.

pbakaus (approved):
**Claude Code, running on Paul's behalf (`@pbakaus`) at his request. AI assistance disclosed per `CLAUDE.md`.**

All three items from the earlier review are addressed, and precisely. Approving.

**Composition.** The two-group precedence is what I asked for: Impeccable patterns govern every path they match, package-manager discovery fills in the rest, and each group's `!` negations stay scoped to their own source. `isSelectableCandidate` also closes a hole I did not name, so the picker no longer offers a folder that would resolve somewhere else. I checked both of the ambiguous cases by hand rather than relying on the tests. With `projectRoots: ["apps"]` over `workspaces: ["apps/*"]`, the picker lists `apps` alone and `--target apps/web` resolves to `apps`. Correct.

**Tests.** The five interaction cases cover exactly the overlaps I flagged, including the broad-pattern-wins-boundary case and the negation-only config.

**Docs.** `config.md` states the precedence rule outright, and it matches the implementation.

**Naming.** `projectRoots` and `readProjectPatterns()` both adopted.

Extra verification, since this sits on the path every project takes:

- A/B'd `context.mjs` against `main` across 18 invocations (workspaces with negation, plain repo, `.impeccable` config with no `projectRoots`; six targets each). Byte-identical output in all 18, so the empty-group path is a real no-op.
- Traversal is contained: `["../outside", "/etc", "docs/*"]` sends `--target ../outside` back to the repo root and lists only `docs/x`.
- Adversarial config values all degrade without crashing: `"."`, `"*"`, `"docs/**"`, a non-array `projectRoots`, junk entries, malformed JSON.
- Clean merge with `main`, full suite green, `bun run build` green, source-only diff.

I refreshed the PR description myself rather than send it back to you for a wording fix. It still described the key as `contextRoots` and carried the naming question as open, so it no longer matched the branch. Code untouched, and there is a note at the top saying so.

Worth flagging for later, not for this PR: `projectRoots` is a new top-level key while the rest of the config lives under `hook` and `detector`. Defensible, since this is neither, but if we ever want a `context` section it is cheaper to decide before the key is public.

Nice work, and thanks for the patience on the second pass.

PR #319, opened 2026-06-29
Add native Cursor plugin via canonical harness paths

harshav167:
## Summary

Adds a native Cursor plugin by generating `.cursor-plugin/plugin.json` and `.cursor-plugin/marketplace.json` from the build. The manifest points at the canonical harness tree already shipped in this repo (`.cursor/skills/`, `.claude/agents/`, `.cursor/hooks.json`) instead of copying a duplicate `plugin-cursor/` payload.

Cursor users can install from the marketplace (`source: "./"`) or symlink the repo locally. The pre-write blocking hook reuses the same `.cursor/hooks.json` path and script as project-level installs.

## Type of change

- [x] Build system / tooling
- [x] Documentation update

## Checklist

- [x] Source files updated in `source/` (`skill/` unchanged; build/transformer/test sources updated)
- [x] `bun run build` ran successfully
- [x] `bun test` passes (`tests/hook-build.test.mjs`, `tests/validate-plugin-versions.test.js`)
- [x] Tested with at least one provider (Cursor — smoke-tested pre-write hook)
- [x] README / DEVELOP.md updated if needed

## Test plan

- [x] `bun run build:release`
- [x] `bun test tests/hook-build.test.mjs`
- [x] `bun test tests/validate-plugin-versions.test.js`
- [ ] Install plugin locally (`ln -sf "$(pwd)" ~/.cursor/plugins/local/impeccable`) and verify skill, agent, and hook under **Customize**


pbakaus:
> Thanks for the PR. Impeccable is moving quickly, and this PR is currently waiting on contributor action.
> 
> It has been open for 10 days. Please address the outstanding review feedback, CI failure, merge conflict, draft state, or explicit maintainer wait request. PRs that are still waiting on contributor action after 14 days open are closed automatically.
> 
> If nothing changes, this PR may be closed on or after 2026-07-13. Happy to reopen when it is ready to continue.

sorry about this aggressive note - the bot is new and didn't realize the merge conflict happened today! I'll be reviewing this PR today.

pbakaus (changes requested):
**This review was produced by Claude Code, running on Paul's behalf (`@pbakaus`) at his request. AI assistance disclosed per `CLAUDE.md`. Paul has the final word on anything here.**

Thanks for this. The core instinct is right, and I want to say that up front: pointing the manifest at the canonical `.cursor/skills/` tree instead of vendoring a duplicate `plugin-cursor/` payload is the correct call, and it avoids a whole second sync surface that would have rotted. `.cursor-plugin/plugin.json` is the right filename and location, `preToolUse` is a valid hook event, `.cursor/hooks.json` is genuinely tracked, and the version-drift tests you added are well built. They are what caught the bug below.

Requesting changes on five things, two of them blocking.

---

## Blocking

### 1. `bun run build` fails on this branch

```
❌ .cursor-plugin/marketplace.json: version "3.8.0" disagrees with .claude-plugin/plugin.json "3.9.1"
❌ .cursor-plugin/plugin.json: version "3.8.0" disagrees with .claude-plugin/plugin.json "3.9.1"
error: script "build" exited with code 1
```

Both manifests are committed at `3.8.0`, and this PR adds them to `validatePluginVersions` in the same change. `main` moved to `3.9.1` since. CI is failing all three test jobs for this reason (`bun run build` runs at `ci.yml:93`). Copilot and Greptile both flagged it too.

`bun run build:release` regenerates them correctly and validation then passes, so the immediate fix is to run it and commit the two files. Please do that as part of item 2 rather than on its own, because on its own it comes straight back.

### 2. `.cursor-plugin` is missing from the generated-output sync, so the drift recurs

This is the part I would most like changed, and it is a one-line fix.

`.github/workflows/sync-generated-output.yml` scopes both of its git calls to an explicit allowlist:

```bash
changes="$(git status --porcelain -- $GENERATED_PATHS)"
git add $GENERATED_PATHS
```

`GENERATED_PATHS` contains `.cursor`, `.claude`, `plugin`, and the rest, but not `.cursor-plugin`. So after this merges:

1. every version bump regenerates `.cursor-plugin/*.json` locally
2. the post-merge sync neither detects nor commits that drift
3. `bun run build` breaks for everyone until somebody hand-fixes it
4. `scripts/release.mjs:121` refuses to release when `build:release` leaves the tree dirty, so **every skill release gets blocked by this**

Regenerating once fixes the symptom and leaves the cause in place. Please add `.cursor-plugin` to `GENERATED_PATHS`.

While you are there: `.cursor-plugin/` is a new generated-but-tracked directory and it is documented as such in zero places. Worth a line in `AGENTS.md`'s tracked-folder enumeration alongside the others, and in the explanatory note in `.gitignore`. PR #373 did the equivalent for `.vibe/skills/`.

---

## Please also address

### 3. `marketplace.json` is invalid against Cursor's published schema, and probably should not exist

I validated both files against Cursor's own schemas in [`cursor/plugins`](https://github.com/cursor/plugins):

| File | Result |
|---|---|
| `.cursor-plugin/plugin.json` | validates cleanly, every key legal |
| `.cursor-plugin/marketplace.json` | 4 illegal keys in the plugin entry |

[`marketplace.schema.json`](https://raw.githubusercontent.com/cursor/plugins/main/schemas/marketplace.schema.json) defines `pluginEntry` with `additionalProperties: false` and exactly three allowed keys: `name`, `source`, `description`. Your entry also carries `version`, `category`, `homepage`, and `tags`. Cursor's own official marketplace manifest uses only the three. The docs table lists more fields than the schema allows, and the schema is the machine-readable one.

Note the knock-on: `collectPluginVersions` reads `plugins[0].version`, which is one of the fields Cursor's schema rejects. The field driving item 1 is not a legal field.

Two more reasons to reconsider the file: `marketplace.json` is documented only for [multi-plugin repositories](https://cursor.com/docs/reference/plugins#multi-plugin-repositories), and `source: "./"` is undocumented and self-referential (it resolves to `./.cursor-plugin/plugin.json`, the same file sitting next to it). Every documented and official example uses a bare subdirectory name.

**Suggestion: drop `.cursor-plugin/marketplace.json`.** A single-plugin repo does not need one. That removes the illegal fields, the self-reference, and one of the two files in item 1 all at once. If you would rather keep it, trim the entry to `name` / `source` / `description` and drop the corresponding validator hook.

### 4. `"agents": "./.claude/agents/"` is the one claim I could not verify, and your own test plan flags it

Cursor genuinely does read Claude-format subagents; `.claude/agents/` is a [documented discovery location](https://cursor.com/docs/agent/subagents). What is not attested anywhere is a **plugin manifest's** `agents` key resolving into a hidden dot-directory, or plugin packaging carrying `.claude/**`. Packagers commonly exclude dot-directories, and no official Cursor plugin does this; they all use `./agents/`.

The last box in your test plan is the one that settles it:

> - [ ] Install plugin locally (`ln -sf "$(pwd)" ~/.cursor/plugins/local/impeccable`) and verify skill, agent, and hook under **Customize**

Could you tick that before we merge? The current tests assert the JSON matches itself, so a plugin that silently fails to load still passes them. Same question applies to the hook: `.cursor/hooks.json` runs `node ".cursor/skills/impeccable/scripts/hook-before-edit.mjs"`, which is workspace-relative and correct for a project install, but Cursor's docs show hook commands relative to the *plugin* root. Please confirm both actually resolve when installed from `~/.cursor/plugins/local/`, and paste what you see.

If `agents` turns out not to resolve from a dot-directory, pointing it at a non-hidden path is the fallback.

### 5. Two small ones

**README option numbering is now `1, 2, 3, 3, 4, 5`.** Inserting the new Option 2 renumbered only the heading directly after it, so there are two Option 3s ("Git Submodule" and "Plugin install"), and the last two should be 5 and 6.

**New hardcoded command count** at `scripts/build.js:834`: `'Design fluency for AI harnesses. 1 skill, 23 commands, ...'`. `generateCounts` validates `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` but not `scripts/build.js` or `.cursor-plugin/*`, so this string goes stale silently the next time a command is added, while the guard correctly forces the `.claude-plugin` files to update. It is also inconsistent inside the same object: `plugins[0].description` derives from `rootManifest.description`, but `metadata.description` is hardcoded. Deriving it matches the rule `CLAUDE.md` sets for the OG image ("read live from `command-metadata.json`, so it never goes stale; don't hardcode it"). Moot if you drop `marketplace.json` per item 3.

---

## Summary of the ask

- add `.cursor-plugin` to `GENERATED_PATHS`, then `bun run build:release` and commit
- drop `.cursor-plugin/marketplace.json` (or trim it to the three legal keys and drop its validator entry)
- document `.cursor-plugin/` in `AGENTS.md` and the `.gitignore` note
- fix the README numbering to 1 through 6
- tick the local-install box and report what loads, especially the agent and the hook

`plugin.json` itself needs no changes. It is correct, and once the sync and marketplace items are sorted this should be a clean merge.

pbakaus:
Thanks for taking the time to contribute. Impeccable is issue-first for outside contributions, and I could not find a relevant linked issue where a maintainer approved or requested this work. I’m closing this for now. Please open or join an issue, wait for explicit approval from @pbakaus or @abdulwahabone, and then open a fresh PR. This policy check was performed automatically by Codex; if it missed existing approval, please point a maintainer to it.

PR #328, opened 2026-07-01
Add codex-grid-background detector rule

pbakaus:
Adds a deterministic detector for the Codex **two-axis grid-line background** tell — the CSS grid overlay built from hairline `linear-gradient(... 1px, transparent 1px)` layers plus a repeating `background-size` cell. The skill-text rule for this landed in 4ac03480; this makes it enforceable by the CLI, browser extension, and site overlay.

## The rule
`codex-grid-background` (`slop`, `gated: 'gpt'`, advisory). Flags a single `background` value containing **two or more** hairline gradient stops (`Npx, transparent Npx`) — one per axis. Off by default; surfaces under `--gpt` / `providers:['gpt']`, consistent with the sibling `repeating-stripes-gradient` codex rule.

## Precision choices (false-positive guards)
- Counts hairline stops **within a single background declaration**, not globally, so two unrelated single-axis ruled lines on one page don't add up to a false flag.
- Matches the `Npx, transparent Npx` stop directly rather than splitting whole gradient layers, because colors like `oklch(96% 0.012 82 / 0.055)` carry nested parens.
- Single-axis ruled lines and plain two-color gradient fades stay silent (both in the fixture pass column).

## Verification
- `bun run test` — 198 pass
- `bun run build` / `build:browser` / `build:extension` — green
- Real-CLI scan of a fixture copy: rule fires under `--gpt`, absent by default
- Confirmed against the canonical Codex snippet (hairline grid + radial spotlights in one `background`) — the two hairline layers are still counted correctly

## Notes
- Rule count bumped 44 -> 45 across `index.astro`, `README.md`, `README.npm.md`
- Skill version intentionally **not** bumped
- Gitignored extension/site bundles regenerate at release time

---

> [!NOTE]
> **Low Risk**
> Additive detector rule behind existing GPT gating; no auth or data paths touched, with fixture coverage for false-positive guards.
> 
> **Overview**
> Adds **`codex-grid-background`**, a new advisory slop rule (gated under `--gpt` / `providers:['gpt']`) that flags decorative two-axis grid backgrounds built from paired hairline `linear-gradient` stops plus a tiled `background-size`.
> 
> The same detection logic is wired into the **CLI static checks** (`checks.mjs`), **browser detector bundle** (`detect-antipatterns-browser.js`), and **antipattern registry**, with guards so only co-located signals in one style block count and hairlines must live in `background` / `background-image` values.
> 
> **Docs and marketing copy** bump the deterministic rule count **44 → 45** (`README.md`, `README.npm.md`, `index.astro`). Tests extend the GPT fixture and expect one hit per gated id when `--gpt` is on.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a334c5a16be51ac07f83cf1577d9b931ea4466b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Impeccable
 
-Design guidance for AI coding agents. 1 skill, 23 commands, live browser iteration, and 44 deterministic detector rules for AI-generated frontend design.
+Design guidance for AI coding agents. 1 skill, 23 commands, live browser iteration, and 45 deterministic detector rules for AI-generated frontend design.
 
 > **Quick start:** From your project root, run `npx impeccable install`, then run `/impeccable init` inside your AI coding tool. Full docs: [impeccable.style](https://impeccable.style).
 
@@ -13,7 +13,7 @@ Every model trained on the same SaaS templates. Skip the guidance and you get th
 Impeccable adds:
 - **One setup flow.** `/impeccable init` writes `PRODUCT.md` and offers `DESIGN.md`, so later commands know the audience, brand/product lane, voice, anti-references, colors, type, and components.
 - **23 commands.** A shared design vocabulary with your AI: `polish`, `audit`, `critique`, `distill`, `animate`, `bolder`, `quieter`, and more.
-- **44 deterministic detector rules** plus LLM-only critique checks. The CLI and browser extension run the deterministic rules with no LLM and no API key.
+- **45 deterministic detector rules** plus LLM-only critique checks. The CLI and browser extension run the deterministic rules with no LLM and no API key.
 
 ## What's Included
 
@@ -341,7 +341,7 @@ npx impeccable ignores add-file "src/legacy/**"
 npx impeccable ignores add-value overused-font Inter --reason "Brand font"
 ```
 
-The detector catches 44 deterministic issues across AI slop (side-tab borders, purple gradients, bounce easing, dark glows) and general design quality (line length, cramped padding, small touch targets, skipped headings, and more).
+The detector catches 45 deterministic issues across AI slop (side-tab borders, purple gradients, bounce easing, dark glows) and general design quality (line length, cramped padding, small touch targets, skipped headings, and more).
 
 By default, `detect` respects the same `.impeccable/config.json` and `.impeccable/config.local.json` detector config as the design hook: `detector.ignoreRules`, `detector.ignoreFiles`, `detector.ignoreValues`, and `detector.designSystem.enabled`. Hook lifecycle settings such as `hook.enabled` only affect automatic hook execution.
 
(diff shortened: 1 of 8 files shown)


pbakaus:
Thanks for the review. Handled both:

**1. Require grid sizing (in-diff) — fixed** in 00e69b60.
Valid catch: two hairline gradients alone are a fixed crosshair, not a grid, and the skill rule itself says "plus \`background-size\`". The detector now scopes to a single style block (a CSS rule body or one inline \`style=\"...\"\`) and requires **both** ≥2 hairline stops **and** a tiling \`background-size\` px cell in that same block. Added a \`pass-crosshair-no-tiling\` fixture case (two hairline gradients, no \`background-size\`) which is now correctly silent; the tiled grid still flags.

**2. Browser detector surfaces gated rule by default (outside-diff P1) — declining, not a regression.**
This is deliberate, pre-existing behavior, documented at \`cli/engine/browser/injected/index.mjs:1461\`: provider-gated rules are intentionally **not** filtered in the browser/overlay/extension path because running every check there is free; gating is purely a CLI output concern applied in the Node engines' \`detect*\` return paths. It applies identically to the four existing gated rules (\`repeating-stripes-gradient\`, \`theater-slop-phrase\`, \`gpt-thin-border-wide-shadow\`, \`image-hover-transform\`). This PR follows that established contract; changing it would alter behavior for those four rules and belongs in a separate, intentional change rather than here.

pbakaus:
Second round handled:

**1. Scope hairlines to backgrounds (in-diff) — fixed** in a334c5a1.
Valid catch. The block matcher now counts hairline stops only inside \`background\` / \`background-image\` declaration values, not the whole rule body, so a hairline in an unrelated property (\`mask-image\`, \`border-image\`) can't stand in for the second axis. Added a \`pass-bg-plus-mask-hairline\` fixture case (one background hairline + \`background-size\` + a \`mask-image\` hairline) which is now correctly silent.

**2. "codex-grid-background does not fire for valid GPT-gated grid backgrounds" (outside-diff P1) — not reproducible; this is a wrong-entry-point artifact.**
The T-Rex harness exercised \`detectText(...)\`, which is the JS/JSX/Vue/Svelte source-text engine. CSS-in-HTML is handled by \`detectHtml\`, where \`checkHtmlPatterns\` runs (same wiring as the four sibling gated rules). The finding does fire under \`providers:['gpt']\` on both requested positive cases:

\`\`\`
grid inline (canonical)   -> [ 'codex-grid-background' ]
grid in <style> rule body -> [ 'codex-grid-background' ]
\`\`\`

and the fixture regression test (\`gpt-tells: with providers:[gpt], each flag case triggers once\`) passes. No change needed.

**3. "Browser detector emits codex-grid-background without GPT provider enabled" (outside-diff P1) — declining again.**
Same as the prior round: provider gating is intentionally CLI-only, documented at \`cli/engine/browser/injected/index.mjs:1461\`, and applies identically to all existing gated rules. Out of scope for this PR.

PR #345, opened 2026-07-05
Add mechanical pre-scan for typeset and layout

### LINKED ISSUE #149: [Bug] Typeset and Layout often need to be run mutliple times to catch basic things

abdulwahabone:
Fixes #149

Follows Paul's suggestion in the issue: scoped detector pre-flight before LLM judgment, plus DESIGN.md font-size validation for the personalized type pass.

Adds `--scope type|layout` to detect, scopes layout/type rules, and wires pre-scan steps into `typeset.md` and `layout.md`.

Tested on my local test repos: `/impeccable typeset` and `/impeccable layout` both ran the mechanical pre-scan.

## Test plan
- [x] `bun run test`
- [x] Manual agent runs on local test repo (typeset + layout pre-scan)

---

> [!NOTE]
> **Low Risk**
> Changes extend the detector and CLI with additive flags and one advisory design-system rule; default unscoped `detect` behavior is unchanged, with tests covering scope parsing and font-size validation.
> 
> **Overview**
> Adds a **mechanical pre-scan** path for `/impeccable typeset` and `/impeccable layout`: parallel sub-agents (or a strict fallback order) run an isolated checklist plus `detect.mjs --json --scope type|layout`, then merge findings before fixes. Verification steps now require evidence and a re-scan until issues are resolved or explicitly accepted.
> 
> **CLI:** `impeccable detect` gains **`--scope type|layout`** (comma-separated), with validation so a bare or unknown scope fails instead of scanning unscoped. Many detector rules are tagged with **`scopes`**; post-scan filtering keeps only matching findings.
> 
> **DESIGN.md:** New advisory rule **`design-system-font-size`** (46 rules total) flags literal **`px`/`rem`** sizes (CSS, `fontSize`, `text-[…]`) that are off the typography ramp; **`em`**, **`clamp()`**, and browser-computed sizes are abstained. Docs and marketing copy bump the rule count to 46.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f3d8edacceb457d8f6a3953bb52f8984f7a785e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pbakaus:
I noticed this doesn't use sub-agents - is there a particular reason for that? If you tested its performance, great, but I think it might perform better if we do something like critique:

```
Spawn two sub-agents (if you need explicit permission by the user to use sub-agents, stop and ask before proceeding). One to run a mechanical pre-flight, and one to analyze the current layout. Sub-agents are critical because one result would bias the other. When you get the results from both, synthesize the findings.
```

abdulwahabone:
> I noticed this doesn't use sub-agents - is there a particular reason for that? If you tested its performance, great, but I think it might perform better if we do something like critique:
> 
> ```
> Spawn two sub-agents (if you need explicit permission by the user to use sub-agents, stop and ask before proceeding). One to run a mechanical pre-flight, and one to analyze the current layout. Sub-agents are critical because one result would bias the other. When you get the results from both, synthesize the findings.
> ```

Thats a good idea, let me add that.
Also, i'll keep this technique in mind.

abdulwahabone:
> I noticed this doesn't use sub-agents - is there a particular reason for that? If you tested its performance, great, but I think it might perform better if we do something like critique:
> 
> ```
> Spawn two sub-agents (if you need explicit permission by the user to use sub-agents, stop and ask before proceeding). One to run a mechanical pre-flight, and one to analyze the current layout. Sub-agents are critical because one result would bias the other. When you get the results from both, synthesize the findings.
> ```
Screenshot 2026-07-09 at 13 59 34

Really like this update, its better and faster now. Tested it with local repos and verified their results with fable

pbakaus:
Nice! Let's get this in!

pbakaus (approved):

Issue #305, opened 2026-06-25
[Bug] PostToolUse hook re-creates .impeccable/ on non-UI edits (cache persisted unconditionally, keyed to session cwd)


steven-pribilinskiy:
### Summary

The PostToolUse hook re-creates the `.impeccable/` directory (writing `hook.cache.json`) after **any** `Edit`/`Write`/`MultiEdit` — including edits to files the hook itself decides to skip, like `.md`/`.json`/`.py`. The extension allow-list only gates *detection*, not the cache write, so a non-UI edit still materializes `.impeccable/` from nothing. Combined with the cache being keyed to the **session cwd** rather than the edited file's project, this makes the folder reappear at an umbrella/multi-project directory and accumulate entries from unrelated projects.

Version: `3.8.0` (latest skill release).

### Repro

1. Launch your agent from a directory that is *not* itself the UI project — e.g. a multi-project umbrella dir (`~/projects`) that is not a git repo and has no `package.json`.
2. `rm -rf .impeccable` in that dir.
3. Edit any non-UI file (a `.md` note, a `.json`, etc.) with Edit/Write.
4. `.impeccable/hook.cache.json` is back, even though nothing UI-related was touched.

### Why it happens

In `skills/impeccable/scripts/hook-lib.mjs` (`runHook`):

- The per-file loop skips a non-allowed extension and `continue`s without recording anything:
  ```js
  if (!ALLOWED_EXTS.has(ext)) {
    lastSkip = 'extension';   // ~line 1454
    continue;
  }
  ```
  (`ALLOWED_EXTS` ~line 46 is `.tsx .jsx .html .htm .vue .svelte .astro .css .scss .sass .less .ts .js` — `.md` is not in it, as intended.)

- But after the loop, the cache is persisted **unconditionally**:
  ```js
  persistCache(projectCwd, cache);   // ~line 1516
  ```
  `persistCache` → `writeCache` → `getCachePath(cwd)` (~line 122, `path.join(cwd, '.impeccable', 'hook.cache.json')`) → `fs.mkdirSync(dirname, { recursive: true })` + `fs.writeFileSync(...)`. So the directory is re-created even when every file in the event was skipped and the cache was never mutated.

[...]

PR #346, opened 2026-07-05
Fix: design hook no longer creates .impeccable/ in unrelated projects

### LINKED ISSUE #344: PostToolUse hook creates .impeccable/ in every project, even ones that never use Impeccable
### Summary
The plugin's `PostToolUse` hook (`hooks/hooks.json` -> `hook.mjs`) runs after every `Edit`/`Write`/`MultiEdit` in **every** project, and writes `.impeccable/hook.cache.json` into the working directory (`hook-lib.mjs`, `fs.mkdirSync(...'.impeccable'...)` / `fs.writeFileSync`). Since the plugin is enabled globally, any repo I edit a file in -- including ones that have nothing to do with UI or Impeccable and where I've never invoked the skill -- gets an `.impeccable/` folder created and repeatedly recreated.

### Impact
- Unwanted, undeclared folder appears in unrelated repos.
- Deleting it does nothing -- the next edit recreates it.
- Shows up as untracked cruft in `git status` unless every project separately gitignores it.

Observed with impeccable v3.7.1 on Claude Code (Windows). The cache is often just `{"version":1,"sessions":{}}` -- empty, so the folder is created even when the hook has nothing to record.

### Request
Only create `.impeccable/` (and write the hook cache/audit files) when the project actually opts into Impeccable -- e.g. the skill has been invoked in this project, or an `.impeccable/config.json` / `DESIGN.md` already exists. If the project has no Impeccable footprint, the hook should detect and exit without creating any files. An early-exit that skips the whole detector when there's no existing `.impeccable/` marker would solve it.

Happy to test a fix.

abdulwahabone:
## Summary

The PostToolUse hook was writing `.impeccable/hook.cache.json` after every edit, even when the file was skipped or the cache never changed. That meant unrelated repos picked up an empty `.impeccable/` folder on the next Markdown or backend edit, and it kept coming back after you deleted it ([#344](https://github.com/pbakaus/impeccable/issues/344)).

This change only persists the cache when there is something worth writing: fresh detector findings, or an edit count update in a project that already has a `.impeccable/` footprint. Skipped files and clean scans in projects that never opted in are now a no-op on disk.

When the agent is launched from a parent folder that is not itself a project, the hook now keys cache and config to the edited file's nearest project root (`.git`, `package.json`, or `.impeccable`) instead of piling everything into the launch directory ([#305](https://github.com/pbakaus/impeccable/issues/305)).

Source-only diff: `skill/scripts/hook-lib.mjs`, `hook-before-edit.mjs`, and tests. No generated harness sync in this PR.

Closes #344, #305

## Test plan

- [x] `node --test tests/hook.test.mjs` (cache gating + umbrella `resolveCacheCwd` cases)
- [x] `bun run test`

**Manual (Codex, local bundle from this branch)**

- [#344](https://github.com/pbakaus/impeccable/issues/344): opened an Astro project with no `.impeccable/` footprint, edited `README.md` and made a trivial clean `.astro` change — no folder appeared. Added an obvious font tell (`Inter` inline) — folder appeared, as expected.
- [#305](https://github.com/pbakaus/impeccable/issues/305): launched Codex from a bare parent directory with the app nested inside (symlink). A Markdown edit left both the parent and the app root clean. A design finding in the app wrote cache under the app’s `.impeccable/`, not the parent.

### DIFF
--- a/skill/scripts/hook-before-edit.mjs
+++ b/skill/scripts/hook-before-edit.mjs
@@ -27,6 +27,7 @@ import {
   readCache,
   readConfig,
   renderTemplate,
+  resolveCacheCwd,
   resolveProjectCwd,
   truthy,
   writeAuditLog,
@@ -379,9 +380,12 @@ async function main() {
     return allow({ skipped: 'stdin-empty' });
   }
 
-  const cwd = resolveProjectCwd(event);
+  const sessionCwd = resolveProjectCwd(event);
   const started = Date.now();
-  const filePath = proposedFilePath(event, cwd);
+  const filePath = proposedFilePath(event, sessionCwd);
+  // Re-key config/cache to the edited file's project root when the session
+  // was launched from a non-project umbrella directory (issue #305).
+  const cwd = resolveCacheCwd(filePath, sessionCwd);
   const audit = {
     harness: 'cursor',
     cwd,
--- a/skill/scripts/hook-lib.mjs
+++ b/skill/scripts/hook-lib.mjs
@@ -10,7 +10,7 @@
  *   truthy(value)
  *   readConfig(cwd) / DEFAULT_CONFIG / getConfigPath(cwd) / getLocalConfigPath(cwd)
  *   normalizeIgnoreValue(value)
- *   readCache(cwd) / persistCache(cwd, cache)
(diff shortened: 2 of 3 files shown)


pbakaus (commented):
LGTM, thanks!

Issue #316, opened 2026-06-27
[Feature] Configurable detector extensions + server-side template support (Blade / Twig / ERB / Handlebars)


LeFrenchie:
## TL;DR

The design hook silently covers **almost nothing** in server-side template stacks (Laravel, Rails, Symfony, .NET, Go templates). Their markup lives in `.blade.php`, `.html.erb`, `.twig`, `.hbs`, etc. — none of which the detector scans. The fix is small: make the scanned extensions configurable and route template files through the existing HTML detector.

## The problem

`ALLOWED_EXTS` in `scripts/hook-lib.mjs` is a hardcoded set:

```js
export const ALLOWED_EXTS = new Set([
  '.tsx', '.jsx', '.html', '.htm', '.vue', '.svelte', '.astro',
  '.css', '.scss', '.sass', '.less', '.ts', '.js',
]);
```

Two things break for template files:

1. **Double extensions are invisible.** The gate uses `path.extname(filePath)`, so `path.extname('Card.blade.php')` returns `.php` — not `.blade.php`. There is no way to express it in the set today.
2. **Even if it passed the gate, it wouldn't route correctly.** The detector only sends `.html`/`.htm` to `detectHtml`; everything else goes to `detectText`. Template markup (Tailwind classes, inline `style=""`, `<style>` blocks) wants the HTML engine.

### Why this matters

In a Laravel + Livewire or a Rails + ERB project, **~100% of the UI** is in these template files. The hook installs cleanly, reports "enabled", and then never fires on the files where the design actually lives. It looks like it works; it doesn't.

### Repro

1. Enable the hook in a Laravel project.
2. Edit a `.blade.php` file with an obvious tell (e.g. gradient text).
3. The hook exits 0 with no finding — the file was skipped at the extension gate.

## Proposed fix

Let `.impeccable/config.json` declare extra extensions. Each entry carries the **one** decision that matters per file: which engine analyzes it (`html` for markup, `text` for JS/TS/CSS):

```jsonc
{
  "detector": {
    "extensions": [
      { "ext": ".blade.php", "engine": "html" },
      { "ext": ".html.erb",  "engine": "html" },
      { "ext": ".twig",      "engine": "html" },
      { "ext": ".hbs",       "engine": "html" }
    ]
  }
}
```

One list, one decision per entry — no second parallel list to keep in sync. `engine` could even default to `html` (the common case for template files), so most configs are just `{ "ext": ".blade.php" }`.

Implementation notes:

[...]

PR #351, opened 2026-07-07
Prefer CLAUDE_PROJECT_DIR / CURSOR_PROJECT_DIR in resolveCacheCwd()

m-tanner:
## Summary

`resolveCacheCwd()` (added in #346 for #305) climbs from the edited file to
the nearest `.git`/`package.json`/`.impeccable` marker when the session cwd
doesn't itself look like a project root. In a monorepo, though, a package
nested under the real project root often has its own `package.json`, so the
session cwd short-circuits *there* before the climb ever runs, and the cache
still scatters across packages instead of converging on the actual project
root.

`envProjectDir()` already special-cases `CURSOR_PROJECT_DIR`, but it's wired
into `normalizeHookEvent()`'s cursor branch and `resolveProjectCwd()` only,
never into `resolveCacheCwd()`. So neither Cursor's env var nor Claude Code's
equivalent (`CLAUDE_PROJECT_DIR`, set on every hook invocation) had any way to
override the cache/config location.

This has `resolveCacheCwd()` consult `envProjectDir()` first, and teaches it
about `CLAUDE_PROJECT_DIR` alongside the existing `CURSOR_PROJECT_DIR` check.
When the harness names the project root **and that directory looks like a real
project root**, trust it outright: it wins even when the session cwd is itself
a marker-bearing directory (the monorepo nested-package case).

### Why the `looksLikeProjectRoot` gate

Claude Code sets `CLAUDE_PROJECT_DIR` on every hook invocation, to the
directory Claude was *launched* from. Trusting it unconditionally would make
the #305 umbrella-climb branch dead code, and in the #305 scenario (a bare
umbrella launch dir with no `.git`/`package.json`/`.impeccable`) it would
scatter the cache back into that umbrella, the exact regression #305 fixed. So
the env root wins only when it is itself a marker-bearing project root:

- Monorepo root (`.git` + `package.json`): trusted, so a nested package with
  its own `package.json` no longer captures the cache (#351).
- Bare umbrella launch dir (no markers): not trusted, so the climb still keys
  the cache to the edited file's nearest project root (#305 preserved).

Source-only diff: `skill/scripts/hook-lib.mjs` and `tests/hook.test.mjs`, same
convention as #346. No generated harness sync in this PR.

## Test plan

- [x] `node --test tests/hook.test.mjs` — 148/149 pass. The one failure
      (`Cursor hook scripts > preToolUse routes configured html-engine
      templates through the HTML engine (issue #316)`) reproduces identically
      on unmodified `main` (verified via `git stash`), unrelated to this
      change.
- [x] Unit tests in `resolveCacheCwd()`'s own describe block: `CLAUDE_PROJECT_DIR`
      and `CURSOR_PROJECT_DIR` (both marker-bearing) win over a
      `package.json`-bearing session cwd, with `CLAUDE_PROJECT_DIR` taking
      precedence when both are set; and a **marker-less** env project dir does
      **not** short-circuit the climb, so an umbrella launch dir still keys to
      the edited file's project root (#305).
- [x] Integration test in `runHook()`'s cache-write-gating describe block: a
      package nested under a monorepo root (each with its own `package.json`)
      still gets its cache pinned to a marker-bearing `CLAUDE_PROJECT_DIR`
      instead of the nested package.

### DIFF
--- a/skill/scripts/hook-lib.mjs
+++ b/skill/scripts/hook-lib.mjs
@@ -143,14 +143,21 @@ function looksLikeProjectRoot(dir) {
   });
 }
 
-// Where `.impeccable/` (cache + config) lives for this event. Normally the
-// session cwd, untouched. But when the agent was launched from an umbrella
-// directory that is not itself a project (no .git, package.json, or
-// .impeccable), key to the edited file's nearest project root instead, so a
-// multi-project launch dir doesn't accumulate a shared cross-project cache
-// (issue #305). Climbing stops at the home dir, falling back to the session
-// cwd when no marker is found.
+// Where `.impeccable/` (cache + config) lives for this event. When the
+// harness tells us the project root directly (CLAUDE_PROJECT_DIR,
+// CURSOR_PROJECT_DIR), trust it outright — it's authoritative even when the
+// session cwd is itself a marker-bearing directory, which matters in
+// monorepos where a package nested under the real project root has its own
+// package.json and would otherwise look like a project root in its own
+// right. Otherwise, key to the session cwd, untouched. But when the agent
+// was launched from an umbrella directory that is not itself a project (no
+// .git, package.json, or .impeccable), key to the edited file's nearest
+// project root instead, so a multi-project launch dir doesn't accumulate a
+// shared cross-project cache (issue #305). Climbing stops at the home dir,
+// falling back to the session cwd when no marker is found.
 export function resolveCacheCwd(primaryFile, sessionCwd) {
+  const envRoot = envProjectDir(null);
+  if (typeof envRoot === 'string' && envRoot) return path.resolve(envRoot);
   const base = path.resolve(sessionCwd || process.cwd());
   if (!primaryFile || typeof primaryFile !== 'string' || hasPathTraversal(primaryFile)) return base;
   if (looksLikeProjectRoot(base)) return base;
@@ -1172,6 +1179,9 @@ export function normalizeHookEvent(event, projectCwd, harness = 'claude') {
 }
 
 function envProjectDir(fallback) {
+  if (typeof process.env.CLAUDE_PROJECT_DIR === 'string' && process.env.CLAUDE_PROJECT_DIR) {
+    return process.env.CLAUDE_PROJECT_DIR;
+  }
   if (typeof process.env.CURSOR_PROJECT_DIR === 'string' && process.env.CURSOR_PROJECT_DIR) {
     return process.env.CURSOR_PROJECT_DIR;
   }
--- a/skill/scripts/hook-lib.mjs
+++ b/skill/scripts/hook-lib.mjs
@@ -144,20 +144,28 @@ function looksLikeProjectRoot(dir) {
 }
 
 // Where `.impeccable/` (cache + config) lives for this event. When the
-// harness tells us the project root directly (CLAUDE_PROJECT_DIR,
-// CURSOR_PROJECT_DIR), trust it outright — it's authoritative even when the
-// session cwd is itself a marker-bearing directory, which matters in
-// monorepos where a package nested under the real project root has its own
-// package.json and would otherwise look like a project root in its own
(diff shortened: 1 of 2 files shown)


m-tanner:
## Cleanup pass — one altitude item worth a look

Ran a `/simplify` review over this diff. The change itself is clean and at the right depth; nothing needs fixing to merge. One structural observation surfaced that's out of scope for this PR but may be worth a follow-up (happy to fold it into this PR if you'd prefer):

**`resolveCacheCwd` and `resolveProjectCwd` now derive "the project root" by opposite env precedence.**
- `resolveProjectCwd` treats env as a *last* resort: `event.cwd || workspace_roots[0] || envProjectDir || fallback`.
- After this change, `resolveCacheCwd` treats the *same* env var as the *highest*-priority signal, short-circuiting before `sessionCwd` is even considered.

Across a single hook pipeline (`resolveProjectCwd(...) → sessionCwd → resolveCacheCwd(file, sessionCwd)`) `envProjectDir` ends up consulted twice, at two different tiers — once as a fallback while computing `sessionCwd`, then again at top priority where it can override the very `sessionCwd` just handed in. It's correct for the monorepo case #351 targets, but a maintainer now has to hold two contradictory mental models of the same concept, and cache location can't be reasoned about from `resolveCacheCwd`'s signature alone.

Related, smaller: adding `CLAUDE_PROJECT_DIR` to the shared `envProjectDir` also (intentionally or not) changes `resolveProjectCwd`, `normalizeGitHubEvent`, and the cursor branch of `normalizeHookEvent`, none of which the PR title scopes. Likely benign since those harnesses usually set `event.cwd`, but the widened blast radius isn't visible from the diff hunk.

**A durable version** would factor a single root-resolution rule (e.g. `projectRoot(event, sessionCwd, primaryFile)`, or an explicit `{ cacheRoot, projectCwd }` pair) so the env-vs-cwd precedence is decided in one documented place. If cache-root and project-cwd genuinely must diverge, that divergence deserves one comment stating why rather than two functions silently ordering the same inputs backwards.

I left it out because it's a cross-function refactor that changes behavior beyond this diff's scope. **If you'd like, I can do it on this PR** — otherwise it's a clean follow-up.

For the record, the smaller items I deliberately *didn't* touch: the `typeof envRoot === 'string' && envRoot` guard (matches the file's house-style idiom, incl. both branches of `envProjectDir`), the expanded comment block (the file favors heavy explanatory comments), and the triplicated env save/restore boilerplate in the new tests (no existing `withEnv` helper to match, and the three sites are heterogeneous enough that a shared helper would be a lateral move).

pbakaus:
Thanks, this is very close. I think the monorepo behavior is right, but I’d like to tighten the architecture slightly before merging.

The main issue is that `resolveCacheCwd()` now reads ambient `process.env` directly and lets a marker-bearing `CLAUDE_PROJECT_DIR` / `CURSOR_PROJECT_DIR` win before checking whether the edited file or session cwd actually belongs to that root. That means an inherited agent env can redirect unrelated hook/test runs into the wrong `.impeccable` cache/config directory.

Codex reproduced that on this branch with:

```sh
CLAUDE_PROJECT_DIR=/Users/paulbakaus/code/impeccable node --test --test-name-pattern "resolveCacheCwd|cache write gating" tests/hook.test.mjs
```

That fails several focused tests because cache resolution jumps to the inherited repo root.

Could we make the env root an explicit, harness-aware input instead? Something like:

- `projectDirFromEnv(harness, env)` returns `CLAUDE_PROJECT_DIR` only for Claude, `CURSOR_PROJECT_DIR` only for Cursor, and nothing for GitHub/Codex unless intentionally supported.
- `resolveCacheCwd(primaryFile, sessionCwd, { envRoot })` only trusts `envRoot` when it looks like a project root and the primary file or session cwd is inside it.
- 
That should preserve the intended monorepo fix while avoiding cross-harness/global-env leakage. The rest of the PR looks good to me.

PR #374, opened 2026-07-16
Fix Live accept for Elixir templates in lib/

heliumbrain:
## What broke

In Phoenix apps, Live often wraps markup that lives in `lib/**` inside a `~H"""` block in a `.ex` file. **Accept** in the browser runs `live-accept.mjs`, which searches the tree for `impeccable-variants-start` markers.

That search only considered a handful of frontend extensions (`.html`, `.jsx`, …) and directories like `src/` and `app/`. It never looked at `.ex` files or `lib/`, so accept returned `Session markers not found` even though the session was valid.

## What changed

- `live-accept.mjs` and `live-wrap.mjs` now include `.ex`, `.heex`, `.exs`, and `.eex`, and search under `lib/` as well.
- A small regression test accepts a variant from `lib/my_app_web/components/layouts.ex`.

## How I checked

```bash
node --test tests/live-accept.test.mjs
```

All tests in that file pass (including the new case).

## Note

Detector hooks recently gained configurable template extensions ([#316](https://github.com/pbakaus/impeccable/issues/316) / [#347](https://github.com/pbakaus/impeccable/pull/347)). Live wrap/accept still uses its own hardcoded list; this PR only extends that list for a common Elixir layout. Happy to fold into a broader Live pass if [#371](https://github.com/pbakaus/impeccable/pull/371) is the better home.

---

> [!NOTE]
> **Medium Risk**
> Touches core Live accept/wrap source discovery and file-extension rules; wrong matches could accept the wrong file or miss sessions, but behavior is heavily tested and scoped to search/resolution.
> 
> **Overview**
> Fixes Live **accept** (and **wrap**) failing on Phoenix projects where session markers live in `lib/**/*.ex` with `~H"""` templates or in `.heex`/`.eex` files—cases the design hook already covered via configurable extensions.
> 
> **Shared template extensions:** New `lib/template-extensions.mjs` owns suffix-based extension matching (including double extensions like `.blade.php`), merges built-in Live markup types (frontend defaults plus `.ex`, `.heex`, `.eex`; **not** `.exs`) with `detector.extensions` from `.impeccable` config, and caches per cwd. `hook-lib.mjs` imports and re-exports `matchConfiguredExtension` from there instead of duplicating logic.
> 
> **Shared source walk:** New `live/source-search.mjs` implements `findSourceFile` with privileged roots including `lib/`, always skipping `.impeccable` (so staged artifacts are not mistaken for real source), optional `skipDirs`/`fileFilter`, and guarded `realpathSync`. `live-wrap.mjs` and `live-accept.mjs` drop their duplicate walkers; accept additionally skips `dist`/`build` (wrap can still hit generated output for `generatedMatch`).
> 
> **Copy-edit paths:** Manual-edit evidence search and commit rollback extensions include Elixir template types.
> 
> CI test suite triggers and new/expanded tests cover template resolution, source search, and accept from `lib/.../layouts.ex`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ad824bee3f627cd179829fc745c8c1e7fdc279a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pbakaus (approved):
thanks for this! I have cleaned it up a bit more and merged in main's latest changes, but this otherwise looks sound.

masterkain:
Follow-up: this fixed the extension list, but Live accept still fails on realistic Phoenix trees, for a second reason in the same code path.

`findSourceFile()` in `scripts/live/source-search.mjs` gives up at `MAX_DEPTH = 5`, counted from each entry in `SOURCE_SEARCH_DIRS`. Starting the walk at `lib`:

```
lib(0)/<app>_web(1)/live(2)/admin(3)/components(4)/pages(5)/   <- last directory scanned
                                                     <page>(6)/thing.ex   <- never read
```

Phoenix apps that group LiveView function components by page land at directory depth 6-7 under `lib`. The marker is never found, so Accept and Discard both return `Session markers not found` — the same symptom this PR set out to fix, from a different cause. The session itself is valid and the wrapper is sitting in the file.

The regression test added here uses `lib/my_app_web/components/layouts.ex`, which is depth 2, so it passes and the deeper layout stays uncovered.

Repro against any Phoenix app with a component nested six or more directories below `lib`:

```js
import { findSourceFile } from './scripts/live/source-search.mjs';
import { resolveLiveTemplateExtensions } from './scripts/lib/template-extensions.mjs';

findSourceFile({
  query: 'a-string-that-exists-in-that-deep-file',
  cwd: process.cwd(),
  extensions: resolveLiveTemplateExtensions(process.cwd()),
});
// -> null, while the same query against a shallow lib/ file resolves fine
```

Confirming it is only the depth: `.ex` / `.heex` / `.eex` are present in `LIVE_TEMPLATE_EXTENSIONS` and `lib` is present in `SOURCE_SEARCH_DIRS`, both verified before looking further.

I patched it locally to:

```js
const MAX_DEPTH = Number(process.env.IMPECCABLE_SOURCE_SEARCH_MAX_DEPTH) || 10;
```

After that, a real Discard returned `{"handled": true, "carbonize": false}` and restored the original markup byte-for-byte. That was the first mechanical accept/discard ever to succeed in that codebase — before the patch every session had to be merged by hand.

Suggestion: raise the default, and/or make it configurable the way the extension list already is. If you want a regression case that would have caught this, nesting the fixture about six directories under `lib` rather than two is enough.

One adjacent note in case it is useful: on a server-rendered HEEx target, replying `--reply <id> done --file <path>` arms the browser's no-HMR fallback, which fetches that path and injects the raw file as HTML. For a `.ex` template that renders the template source into the page — un-expanded `{...}` interpolation, `:for` never iterated, attributes mangled — and it survives a reload because the client re-applies the published file on re-attach. Omitting `--file` avoids it entirely, since Phoenix's code reloader already puts the rendered variants in the DOM where the MutationObserver finds them. Might be worth a note in the docs, or skipping the fallback for extensions whose markup is not directly servable.

PR #396, opened 2026-07-21
fix(live-server): restrict CORS to loopback origins so /live.js no longer leaks the session token (#304)

### LINKED ISSUE #304: [P1][security] live-server /live.js leaks bearer token via unauthenticated CORS-readable endpoint
**Version:** impeccable@3.1.0 (also on HEAD)

**File:** `.claude/skills/impeccable/scripts/live-server.mjs` lines 388-414

```js
return (req, res) => {
  const url = new URL(req.url, `http://localhost:${state.port}`);
  res.setHeader('Access-Control-Allow-Origin', '*');
  res.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS');
  res.setHeader('Access-Control-Allow-Headers', 'Content-Type');
  if (req.method === 'OPTIONS') { res.writeHead(204); res.end(); return; }
  const p = url.pathname;
  if (p === '/live.js') {
    // ...
    const body = assembleLiveBrowserScript({
      token: state.token,
      port: state.port,
      // ...
    });
```

**Issue.** Every other route under the live server (`/source`, `/events`, manual-edit endpoints) checks `token === state.token` and returns 401 on mismatch. But `/live.js` returns a JavaScript bundle with `state.token` embedded inline, without any auth check, and the handler sets `Access-Control-Allow-Origin: *` for all responses.

**Impact.** While the live server is running on the user's machine, any web page (or local process — a browser tab visiting an attacker-controlled site, a stray service worker, an `iframe`, etc.) that knows or probes the default localhost port can fetch `/live.js`, regex out the embedded token, and then make authenticated requests against `/source`, `/events`, and the manual-edit routes against the user's project. The token-gated authorization on the rest of the server is undermined.

The default port + the `*` CORS policy + the unauthenticated token-bearing script combine to make this a real local-network / local-browser attack surface, not a theoretical one. Mitigations like "localhost-only listen" don't help: the browser running on the same machine that hosts the live server is exactly the attack vector.

**Fix options:**

1. **Token-gate `/live.js` too**: require `?token=...` on the script URL itself (the live mode's bootstrap can include the token in the script-tag src). Cleanest mitigation.
2. **Drop the token from the served script**: serve a static-ish bootstrap that POSTs to a server-side handshake to *acquire* a per-session token. Larger refactor.
3. **Tighten CORS**: replace `Access-Control-Allow-Origin: *` with the live mode's known origin (the page being driven). Doesn't fix the same-origin case but reduces blast radius.

Surfaced by codex review against a real install. Wanted to file it so future installs aren't silently exposed.

onydigitalmarketing-ux:
Fixes #304.

## Problem

`live-server.mjs` embeds the session bearer token in the body of `/live.js`, but served every response with `Access-Control-Allow-Origin: *`. During an active live session, any page open in the same browser could `fetch('http://localhost:<port>/live.js')` cross-origin, read the token out of the response, and then drive the token-protected endpoints (`/events`, `/poll`, `/source`, `/annotation`, `/status`, ...).

As @abdulwahabone noted on the issue, the blast radius is bounded — it only bites during a live session and only if a malicious tab is open in the same browser and probes the port — so this is more P3 than P1. It's still a real token disclosure with a small, self-contained fix.

## Fix

Reflect the request `Origin` **only when it is loopback** (`localhost` / `127.0.0.1` / `[::1]`, any port) and add `Vary: Origin`. The server always binds `127.0.0.1`, so the dev-server tab on another loopback port is the only legitimate cross-origin caller; it keeps working. A non-loopback page now receives no `Access-Control-Allow-Origin` header at all, so the browser blocks it from reading the token-bearing response.

```js
const requestOrigin = req.headers.origin;
if (isLoopbackOrigin(requestOrigin)) {
  res.setHeader('Access-Control-Allow-Origin', requestOrigin);
  res.setHeader('Vary', 'Origin');
}
```

This is a deliberately minimal change (no handshake/bootstrap rework) — it maps to fix option 3 in the issue.

## Tests

Adds a regression test in `tests/live-server.test.mjs`:
- a non-loopback `Origin` (`https://evil.example.com`) receives **no** `Access-Control-Allow-Origin` header;
- a loopback origin (`http://localhost:5173`) is reflected **exactly** (never `*`) with `Vary: Origin`.

## Notes for reviewers

The edit was made to the canonical source `skill/scripts/live-server.mjs` and propagated to all 14 generated per-provider copies so the tree stays consistent — hence the wide-but-identical diff. Happy to adjust the loopback matcher or split the generated-copy sync if you'd prefer it handled by the build step instead.

---

> [!NOTE]
> **Medium Risk**
> Security fix on local auth token exposure; behavior change may break non-loopback cross-origin dev setups, but intended scope is loopback-only callers.
> 
> **Overview**
> **Tightens CORS on the live dev server** so cross-origin pages in the same browser can no longer read `/live.js`, which embeds the session bearer token.
> 
> Instead of `Access-Control-Allow-Origin: *`, responses now set `Access-Control-Allow-Origin` only when the request `Origin` matches loopback (`localhost`, any `127.*` address, or `[::1]`), with `Vary: Origin`. Legitimate dev-server tabs on another local port keep working; remote or arbitrary origins get no CORS header and cannot read the script body.
> 
> The same change is applied across all mirrored `live-server.mjs` skill bundles. **`tests/live-server.test.mjs`** adds a regression test for evil vs loopback origins (including `127.0.0.2`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0814df7a94d5927ed15d32884ac309fdbda0d608. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/skills/impeccable/scripts/live-server.mjs
+++ b/.agents/skills/impeccable/scripts/live-server.mjs
@@ -85,6 +85,12 @@ async function findOpenPort(start = 8400) {
   });
 }
 
+const LOOPBACK_ORIGIN_RE = /^https?:\/\/(localhost|127\.0\.0\.1|\[::1\])(:\d+)?$/i;
+
+function isLoopbackOrigin(origin) {
+  return typeof origin === 'string' && LOOPBACK_ORIGIN_RE.test(origin);
+}
+
 // ---------------------------------------------------------------------------
 // Session state
 // ---------------------------------------------------------------------------
@@ -630,7 +636,16 @@ function statOrNull(filePath) {
 function createRequestHandler({ detectScript, liveScriptParts }) {
   return (req, res) => {
     const url = new URL(req.url, `http://localhost:${state.port}`);
-    res.setHeader('Access-Control-Allow-Origin', '*');
+    // Reflect only loopback origins. This server always binds 127.0.0.1, so
+    // the dev-server tab (a different loopback port) is the only legitimate
+    // cross-origin caller. A wildcard here let any page open in the same
+    // browser — including a malicious one probing local ports — fetch
+    // /live.js and read the bearer token embedded in its body.
+    const requestOrigin = req.headers.origin;
+    if (isLoopbackOrigin(requestOrigin)) {
+      res.setHeader('Access-Control-Allow-Origin', requestOrigin);
+      res.setHeader('Vary', 'Origin');
+    }
     res.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS');
     res.setHeader('Access-Control-Allow-Headers', 'Content-Type');
     if (req.method === 'OPTIONS') { res.writeHead(204); res.end(); return; }
--- a/.agents/skills/impeccable/scripts/live-server.mjs
+++ b/.agents/skills/impeccable/scripts/live-server.mjs
@@ -85,7 +85,10 @@ async function findOpenPort(start = 8400) {
   });
 }
 
-const LOOPBACK_ORIGIN_RE = /^https?:\/\/(localhost|127\.0\.0\.1|\[::1\])(:\d+)?$/i;
+// Loopback origins only. The whole 127.0.0.0/8 block is loopback (not just
+// 127.0.0.1), so a dev page on e.g. 127.0.0.2 is still same-machine and must
+// be allowed; localhost and the IPv6 loopback [::1] round out the set.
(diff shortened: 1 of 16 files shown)


abdulwahabone:
Closing this, because #304 was fixed in https://github.com/pbakaus/impeccable/issues/304#issuecomment-5075272743

If you can still reproduce the issue, please comment on #304

PR #425, opened 2026-07-27
fix(detect): use system Chrome on Windows to stop GPU crash-loop window (#372)

### LINKED ISSUE #372: [Bug] `impeccable detect <url>` opens a visible browser or black popup window
## What happened?

Running `impeccable detect` against a live URL causes a visible browser-like or black pop-up window to appear.

For example:

```bash
impeccable detect https://example.com
```

The window may appear briefly while the page is being scanned. Since URL detection is expected to use a headless browser, no browser or additional window should be visible.

## Steps to reproduce

1. Open a terminal.
2. Run:

   ```bash
   impeccable detect https://example.com
   ```
3. Observe that a browser-like or black pop-up window appears while the command is running.

## Expected behavior

The live page should be scanned entirely in headless mode, without opening or briefly displaying any browser, console, or black pop-up window.

## Provider & environment

* **Provider** (Cursor / Claude Code / Gemini CLI / Codex / Copilot / Kiro / OpenCode): Any
* **Provider version**: Any
* **OS**: Windows 11 25H2
* **Node.js version**: v24.18.0
* **Impeccable version**: 3.2.1

## Additional context

Command used:

```bash
impeccable detect https://example.com
```

The scan otherwise appears to run, but a visible browser-like or black window is displayed during the process.

Screenshots or a screen recording can be added here if needed.

## Screenshot

Image

vinaypokharkar:
## What this fixes

On Windows, running `npx impeccable detect <url>` pops up a black window that hangs around for the whole scan. It looks alarming (a random window appearing during a CLI command made me wonder if something was wrong security-wise), so I dug into where it comes from.

Fixes #372.

## What's actually happening

Puppeteer downloads its own copy of Chrome into a user-writable cache folder. On Windows that path isn't trusted the same way an installed app is, so when Chrome tries to spin up its GPU process for compositing, that process gets denied (`STATUS_ACCESS_DENIED`) by the GPU sandbox / security software. Chrome doesn't give up: it keeps relaunching the GPU process in a loop, and every relaunch briefly flashes a compositor surface on screen. That flashing surface is the black window people see. It isn't a real app window (not in Alt+Tab, not clickable, not focusable), just a side effect of the crash loop.

The system-installed Chrome doesn't have this problem, because it runs from a trusted location and its GPU process is allowed. I confirmed this from `chrome://gpu`: the installed Chrome reports a healthy GPU with a crash count of 0, while the bundled build crash-loops.

## The workaround

The fix I landed on is small: on Windows, launch the system-installed Chrome via Puppeteer's `channel: 'chrome'` instead of the bundled build. If no system Chrome is found, it silently falls back to the bundled browser so nothing breaks.

Details:

- **Windows only.** Mac and Linux don't have this bug, so they keep using the pinned bundled Chrome for consistent measurements across machines. I didn't want to change behavior for people who aren't affected.
- **Contrast measurement is unchanged.** Both browsers use the same hardware GPU, so the detector's pixel/contrast readings come out identical. I checked the fixture output before and after and the findings match.
- **Graceful fallback.** If `channel: 'chrome'` can't resolve an installed Chrome, it drops back to the bundled build.

While I was in there I also noticed one launch path had `headless: true` hardcoded, so a caller-supplied `headless` option was being ignored. That's fixed too.

## Testing

Added `tests/detect-url-launch.test.mjs` covering the launch strategy with an injected fake Puppeteer and a stubbed `process.platform`:

- Windows prefers `channel: 'chrome'`
- Windows falls back to bundled when system Chrome is unavailable
- non-Windows uses the bundled build directly and never attempts the channel

Verified locally on the affected Windows machine: the black window is gone, the page loads in ~110ms, GPU crash count is 0, and the detector findings are unchanged.

---

This change was prepared with AI assistance.

---

> [!NOTE]
> **Low Risk**
> Scoped to browser launch on Windows with fallback; non-Windows behavior unchanged aside from shared helper and headless wiring.
> 
> **Overview**
> Fixes the **black flashing window** during `detect <url>` on Windows (#372) by centralizing Puppeteer startup in a new **`launchBrowser`** helper.
> 
> On **win32**, it tries **system Chrome** (`channel: 'chrome'`) first to avoid bundled Chrome’s GPU sandbox crash-loop, then **falls back** to the pinned bundled build if channel launch fails. If both fail, the bundled error is thrown with the channel error attached as **`cause`**. **Linux/macOS** still launch bundled Chrome only so measurement stays consistent.
> 
> **`detectUrl`** and **`createBrowserDetector`** now call `launchBrowser` instead of `puppeteer.default.launch` directly; **`detectUrl`** also honors **`options.headless`** instead of always forcing headless.
> 
> Adds **`tests/detect-url-launch.test.mjs`** (fake Puppeteer + stubbed `process.platform`) and registers it in the **detector** test suite.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9c0fc887bc105cd8037bbcb4bd186e695c49e80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ramhaidar:
thought it was merged and fixed in [CLI 3.4.0](https://github.com/pbakaus/impeccable/releases/tag/cli-v3.4.0) 😭😭😭

pbakaus (approved):
thanks, I can't validate this but it looks reasonable! let's give it a try.

PR #455, opened 2026-07-30
fix(live): scan JSX comment variant markers; never whole-file-inject raw JSX (#454)

### LINKED ISSUE #454: [Bug] Live mode: injectVariantsFromSource renders raw JSX as text for React targets (JSX comment markers never match HTML comment scan)
## What happens

On a Vite + React target, when the browser misses/loses the HMR update after `--reply done --file src/App.jsx` (e.g. Fast Refresh does a full reload because the edited file also exports non-component helpers), the recovery path `injectVariantsFromSource()` fetches the raw source via the helper's `/source` endpoint and injects it with DOMParser. The user then sees literal JSX source on the page: `{liste.length}`, `{eksikler.length ? ... : "Yok"}`, and even the marker text `impeccable-variants-start <id> */}` rendered as content.

Screenshot symptom: stat cards showing raw `{expr}` bodies instead of values (happy to attach).

## Root cause (two stacked problems)

1. **Marker mismatch.** `injectVariantsFromSource()` scans for HTML comment markers:
   ```js
   const startMark = '<!-- impeccable-variants-start ' + sessionId + ' -->';
   ```
   But in a JSX file the wrapper markers are necessarily JSX comments (`{/* impeccable-variants-start ... */}`) — an HTML comment is invalid inside a JSX element tree. `indexOf` returns -1, so the code falls through to `block = html` and injects the **entire raw file**.

2. **Raw JSX is not renderable.** Even if the marker scan matched, DOMParser-parsing a JSX variant block only works when the variant contains zero JSX expressions. The inline comment says "Full-file parse works for HTML/JSX", but any `{expression}`, `className=`, `style={{...}}` or self-closing quirk renders as literal text/attributes. So the no-HMR fallback is structurally broken for React targets whose variants use live data from the surrounding scope — which SKILL guidance encourages (real data over fabricated).

## Repro

1. Vite 6 + React SPA, app in a subdir (`panel/`), helper started at repo root.
2. Pick an element whose source block lives inside a component with sibling helper exports (Fast Refresh → full reload on edit).
3. Agent writes wrapper + 4 variants (JSX expressions inside), replies `done --file panel/src/App.jsx`.
4. Reload the page mid-cycle. Browser enters the completed-source fallback and injects raw JSX as text.

## Suggested fixes

- In `injectVariantsFromSource`, also scan for the JSX marker form (`{/* impeccable-variants-start ID */}`) before falling back to whole-file injection — and **never** inject the whole file for `.jsx/.tsx` paths; entering recovery (`variant_anchor_missing`) would be strictly better than rendering source text.
- Longer term: React/JSX needs the same component-preview path Svelte got (#150 / #433 direction) — compile variants through the dev server (e.g. `/@fs` import of a temp module) instead of raw-text injection.

## Environment

- impeccable 4.0.2 (Claude Code plugin), Vite 6, React 18, macOS.

kticoder:
Fixes #454.

## Problem

`injectVariantsFromSource()` scans only for HTML comment markers, but `live-wrap.mjs` necessarily emits **JSX comment markers** (`{/* impeccable-variants-start ID */}`) inside JSX/TSX wrappers (line ~322; an HTML comment is invalid in a JSX element tree). On React targets the scan always misses, so the no-HMR fallback injects the **entire raw source file**, rendering `{expressions}` and marker text as literal page content.

## Fix (kept minimal)

1. Scan both marker syntaxes (`<!-- -->` and `{/* */}`) before falling back.
2. For `.jsx/.tsx` sources with no markers, treat it like a missing wrapper (existing recovery/orphan path) instead of whole-file injection — raw JSX can never render correctly, so recovery is strictly better than showing source text. HTML behavior is unchanged.
3. `normalizeSourceFallbackBlock`: strip JSX comments so `{/* Original */}` etc. don't render as text in the DOMParser preview.

## Testing

- `node --check` passes.
- Marker-scan logic unit-tested standalone: HTML markers (unchanged), JSX markers (now found), JSX-no-marker (empty block → recovery), HTML-no-marker (whole file, unchanged), `.tsx` covered.
- Reproduced the original symptom on a Vite 6 + React 18 app (raw `{liste.length}` on screen) before the patch; described in #454.

Not run: `bun run build` / live-e2e (no bun locally) — happy to iterate if CI flags the generated outputs.

---

> [!NOTE]
> **Medium Risk**
> Changes the live cycling injection path for JSX/React, but behavior is narrowly scoped to marker parsing and adoption with existing recovery retries preserved.
> 
> **Overview**
> Fixes **React/JSX live variant injection** when the no-HMR source fallback runs: the browser no longer treats raw `.jsx`/`.tsx` as injectable HTML or misses JSX-only variant markers.
> 
> For JSX/TSX, **`injectVariantsFromSource`** now **adopts an existing live DOM wrapper** (after Vite reload / Fast Refresh) instead of cloning a `DOMParser` preview, which previously showed `{expressions}` as literal text and broke React’s tree (#454).
> 
> When fetching source, variant blocks are found via **both** `<!-- ... -->` and `{/* ... */}` markers; JSX blocks get a **synthesized wrapper** when markers sit inside the element tree. **Unmarked JSX** skips whole-file injection and uses the existing empty/orphan recovery path; HTML unmarked behavior is unchanged. **`normalizeSourceFallbackBlock`** also strips JSX comments from DOMParser previews so comment markers do not render as visible text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88d501c644ea799083c8a95894f1f356e8694d35. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skill/scripts/live-browser.js
+++ b/skill/scripts/live-browser.js
@@ -6210,14 +6210,34 @@
         const parser = new DOMParser();
         let srcWrapper = null;
 
-        // Full-file parse works for HTML/JSX; Astro/Vue sources need marker extraction.
-        const startMark = '<!-- impeccable-variants-start ' + sessionId + ' -->';
-        const endMark = '<!-- impeccable-variants-end ' + sessionId + ' -->';
-        const startIdx = html.indexOf(startMark);
-        const endIdx = html.indexOf(endMark);
-        const block = startIdx !== -1 && endIdx !== -1 && endIdx > startIdx
-          ? html.slice(startIdx + startMark.length, endIdx).trim()
-          : html;
+        // Marker comments differ by source syntax: HTML/Astro/Vue wrappers use
+        // <!-- ... -->, while JSX/TSX wrappers must use {/* ... */} (an HTML
+        // comment is invalid inside a JSX element tree), so scan for both.
+        // Never fall back to whole-file injection for JSX: raw JSX renders
+        // {expressions} and marker text as literal page content (#454).
+        const markerPairs = [
+          { open: '<!--', close: '-->' },
+          { open: '{/*', close: '*/}' },
+        ];
+        let block = null;
+        for (const pair of markerPairs) {
+          const startMark = pair.open + ' impeccable-variants-start ' + sessionId + ' ' + pair.close;
+          const endMark = pair.open + ' impeccable-variants-end ' + sessionId + ' ' + pair.close;
+          const startIdx = html.indexOf(startMark);
+          const endIdx = html.indexOf(endMark);
+          if (startIdx !== -1 && endIdx > startIdx) {
+            block = html.slice(startIdx + startMark.length, endIdx).trim();
+            break;
+          }
+        }
+        if (block == null) {
+          if (/\.[cm]?[jt]sx$/i.test(String(filePath || ''))) {
+            console.warn('[impeccable] JSX source has no variant markers; skipping raw-source injection (#454).');
+            block = '';
+          } else {
+            block = html;
+          }
+        }
         const doc = parser.parseFromString(normalizeSourceFallbackBlock(block, filePath), 'text/html');
         srcWrapper = doc.querySelector('[data-impeccable-variants="' + sessionId + '"]');
         if (!srcWrapper) {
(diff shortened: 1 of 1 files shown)


pbakaus:
Thanks for taking the time to contribute. This PR links issue #454, but I could not find explicit authorization from @pbakaus or @abdulwahabone for this implementation. This repository requires maintainer approval before implementation, so I’m closing the PR for now. This check was performed automatically by Codex; if this classification is mistaken, a maintainer can correct it.

abdulwahabone:
Reopening after approval

Issue #452, opened 2026-07-30
Windows: installed Codex hook guard uses POSIX shell syntax


PatrickSys:
### Summary

On Windows, `npx impeccable install` writes Codex hook commands using POSIX shell syntax:

```sh
[ ! -f "C:\Users\<user>\.agents\skills\impeccable\scripts\hook.mjs" ] || node "C:\Users\<user>\.agents\skills\impeccable\scripts\hook.mjs"
```

Codex 0.146.0 runs hook commands through `COMSPEC`, normally `cmd.exe /C`. `cmd.exe` does not understand `[ ! -f ... ]`, so the guard fails before Impeccable's hook can run correctly.

### Reproduction

1. On Windows, run `npx impeccable install`.
2. Install Impeccable for Codex at user/global scope.
3. Trigger an applicable edit or the Stop hook.

Running the generated command with `cmd.exe /C` shows the same problem:

```text
'[' is not recognized as an internal or external command,
operable program or batch file.
```

Because `cmd.exe` does understand `||`, it then tries to run Node even when the hook file is missing, which can also produce `MODULE_NOT_FOUND`.

### Expected behavior

The generated Codex manifest should keep its current `command` for macOS and Linux and use Codex's supported `commandWindows` field for Windows. A Windows guard can use `if exist` and should keep the hook's exit code.

### Likely source

In Impeccable 3.5.0, `cli/bin/commands/skills.mjs` contains:

```js
function guardHookCommand(quotedPath) {
  return `[ ! -f ${quotedPath} ] || node ${quotedPath}`;
}
```

That guard is also written into Codex's `.codex/hooks.json`.

Codex source for this version:

- [`commandWindows` is selected on Windows](https://github.com/openai/codex/blob/rust-v0.146.0/codex-rs/hooks/src/engine/discovery.rs)
- [the default Windows shell is `COMSPEC` / `cmd.exe /C`](https://github.com/openai/codex/blob/rust-v0.146.0/codex-rs/hooks/src/engine/command_runner.rs)

This may originate from the guard introduced around #400, but #400 concerns Claude Code Stop-hook re-entry rather than Windows/Codex command parsing. It is also distinct from #445.

### Environment

- OS: Microsoft Windows NT 10.0.26200.0
- Node.js: 22.22.2
- Codex CLI: 0.146.0
- Impeccable npm package: 3.5.0

_AI assistance was used to reproduce, investigate, and write up this report._

abdulwahabone:
[...]

Issue #445, opened 2026-07-29
[Bug] Node version probe fails under Volta on Windows, disabling the hook


rlorenzo:
## What happened?

The Node guard added in #423 runs `node -e "process.exit(parseInt(process.versions.node,10)>=22?0:1)"`. On Windows, Volta's `node` shim executes through `cmd /C`, which reads the `>` as output redirection. The probe fails with `The filename, directory name, or volume label syntax is incorrect` and exits 1 before node runs, so the wrapper concludes there is no usable Node and `hook.mjs` never runs.

The hook is then dead on every PostToolUse and Stop, in every project, on a supported Node 24. The `2>/dev/null` hides the error, and the `~/.impeccable/node-unsupported` marker means the incorrect notice appears once and never again.

## Steps to reproduce

1. On Windows, use Volta to provide Node 24, so `node` resolves to the shim at `C:\Program Files\Volta\node.exe`.
2. Run the probe from the generated hook command, with stderr unmuted:

```
$ node --version
v24.16.0

$ node -e "process.exit(parseInt(process.versions.node,10)>=22?0:1)"
The filename, directory name, or volume label syntax is incorrect.
exit=1
```

3. Edit a UI file in Claude Code. The notice appears once, and the detector never runs again.

## Expected behavior

The probe exits 0 on Node 22/24, and the hook runs.

## Provider & environment

- **Provider**: Claude Code
- **Provider version**: 2.1.220, impeccable plugin 4.0.3
- **OS**: Windows 11 Enterprise 22621, Volta 2.0.2, Node 24.16.0. PowerShell 7.

## Additional context

Volta's Windows shims exec through `cmd /C`, which re-parses the argument list and claims `<`, `>`, and newlines for itself. Upstream: volta-cli/volta#1791, open since July 2024, where the maintainer confirms `cmd.exe` is the cause.

Fix in `scripts/lib/transformers/hooks.js`:

```diff
-const NODE_PROBE = `node -e "process.exit(parseInt(process.versions.node,10)>=${NODE_MAJOR_FLOOR}?0:1)" 2>/dev/null`;
+const NODE_PROBE = `node -e "process.exit(Math.min(parseInt(process.versions.node,10),${NODE_MAJOR_FLOOR})===${NODE_MAJOR_FLOOR}?0:1)" 2>/dev/null`;
```

[...]

PR #458, opened 2026-07-31
Fix: keep the node runtime probe clear of cmd.exe metacharacters

rlorenzo:
- Linked issue: #445 (requested by @abdulwahabone in [this comment](https://github.com/pbakaus/impeccable/issues/445#issuecomment-5128074595))
- Contributor status:
  - [ ] I am `pbakaus` or `abdulwahabone`
  - [x] A maintainer approved this PR in the linked issue

Adjacent to but distinct from #452 and #453: that pair fixes the POSIX `[ ! -f ... ]` guard the installer writes in `cli/bin/commands/skills.mjs`. This fixes the `>=` in the `node -e` probe emitted by `scripts/lib/transformers/hooks.js`. No shared code, no overlapping files.

## Summary

The Node guard from #423 probes with `>=`:

```
node -e "process.exit(parseInt(process.versions.node,10)>=22?0:1)"
```

Volta's Windows shims exec through `cmd /C`, which claims the `>` as redirection. The probe fails with `The filename, directory name, or volume label syntax is incorrect` and exits 1 before node runs, so the guard reports no usable Node and `hook.mjs` never runs. `2>/dev/null` hides the error and the `node-unsupported` marker shows the notice only once, so the hook is silently dead on every PostToolUse and Stop. Upstream: volta-cli/volta#1791.

**Fix.** Clamp instead of compare. Same floor test, same ES5-only syntax, no character cmd.exe can claim:

```
node -e "process.exit(Math.min(parseInt(process.versions.node,10),22)===22?0:1)"
```

**Verification.** Each version pinned with `volta pin`, so `node` resolves through the shim:

| Node | old `>=` | new clamp |
| --- | --- | --- |
| 24.16.0 | exit 1, cmd.exe parse error | exit 0 |
| 24.14.0 | exit 1, cmd.exe parse error | exit 0 |
| 22.18.0 | exit 1, cmd.exe parse error | exit 0 |
| 20.6.1 | exit 1, cmd.exe parse error | exit 1, silent |

The 20.6.1 row shows the old probe returned the right answer for the wrong reason: it never evaluated the version at all.

`tests/hook-build.test.mjs` gains one assertion that the probe carries no character cmd.exe re-parses. The suite already asserts every manifest command contains `NODE_PROBE`, so that covers all seven builders.

The six regenerated manifests are included because the generated command string is the artifact being fixed, and `hook-build.test.mjs` asserts committed manifests match the builders.

## Type of change

- [ ] New command
- [ ] New / updated skill reference
- [ ] New anti-pattern guidance
- [x] Bug fix
- [ ] Documentation update
- [ ] Build system / tooling
- [ ] Other: 

## Checklist

- [x] Source files updated in `source/` <sub>(no `source/` dir in the repo today; the source change is `scripts/lib/transformers/hooks.js`)</sub>
- [x] `bun run build` ran successfully
- [x] `bun test` passes <sub>(green in CI on Node 22.12.0 and 24; a local Windows run has pre-existing symlink-EPERM and path-separator failures that reproduce identically on unmodified `main`)</sub>
- [x] Tested with at least one provider (Cursor / Claude Code / Gemini CLI / Codex / Copilot / Grok Build / Kiro / OpenCode / Qoder / Mistral Vibe) <sub>(Claude Code: ran the generated PostToolUse command end-to-end with `CLAUDE_PROJECT_DIR` set; probe passed and `hook.mjs` executed)</sub>
- [x] README / DEVELOP.md updated if needed
- [x] I reviewed the full diff myself before requesting human review
- [x] I disclosed any AI assistance in this PR and related commits/comments, or no AI assistance was used

---

> [!NOTE]
> **Low Risk**
> Localized change to hook shell command strings and build output; same version floor semantics, with lower risk on Windows Volta/cmd paths.
> 
> **Overview**
> Fixes **silent hook failure on Windows** when Node is launched through Volta shims (and similar `cmd /C` paths): the runtime guard’s `node -e` probe used `>=`, which **cmd.exe treats as redirection**, so the probe exited before Node ran and manifests behaved as if no supported Node was on PATH.
> 
> The probe in `scripts/lib/transformers/hooks.js` now uses a **clamp** (`Math.min(...)===22`) with the same Node 22 floor and ES5-only syntax, but **without `<`, `>`, or newlines** in the one-line `-e` payload. All **six regenerated hook manifests** (Claude, Codex, Cursor, GitHub, Grok, plugin) pick up the new command string.
> 
> `tests/hook-build.test.mjs` updates the shared `NODE_PROBE` expectation and adds a test that the probe contains no cmd.exe-unsafe characters; existing checks still require every manifest command to include that probe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c5e38b83f40578a9665d347730ab0f5c619a18e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -7,7 +7,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "[ ! -f \"${CLAUDE_PROJECT_DIR}/.claude/skills/impeccable/scripts/hook.mjs\" ] || ! { node -e \"process.exit(parseInt(process.versions.node,10)>=22?0:1)\" 2>/dev/null || { D=\"$HOME/.impeccable\"; [ -f \"$D/node-unsupported\" ] || { mkdir -p \"$D\" 2>/dev/null && : > \"$D/node-unsupported\" 2>/dev/null && printf '%s' '{\"systemMessage\":\"The impeccable design hook is not running: no Node 22 or newer on PATH. Install one, or remove the impeccable hook from your harness settings.\"}'; }; exit 0; }; } || node \"${CLAUDE_PROJECT_DIR}/.claude/skills/impeccable/scripts/hook.mjs\"",
+            "command": "[ ! -f \"${CLAUDE_PROJECT_DIR}/.claude/skills/impeccable/scripts/hook.mjs\" ] || ! { node -e \"process.exit(Math.min(parseInt(process.versions.node,10),22)===22?0:1)\" 2>/dev/null || { D=\"$HOME/.impeccable\"; [ -f \"$D/node-unsupported\" ] || { mkdir -p \"$D\" 2>/dev/null && : > \"$D/node-unsupported\" 2>/dev/null && printf '%s' '{\"systemMessage\":\"The impeccable design hook is not running: no Node 22 or newer on PATH. Install one, or remove the impeccable hook from your harness settings.\"}'; }; exit 0; }; } || node \"${CLAUDE_PROJECT_DIR}/.claude/skills/impeccable/scripts/hook.mjs\"",
             "timeout": 5,
             "statusMessage": "Checking UI changes"
           }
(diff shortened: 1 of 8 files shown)


pbakaus (approved):
LGTM

Issue #475, opened 2026-08-01
Bundle served at /api/download/bundle/universal omits reference/degraded/ shipped in the skill-v4.0.4 tag (dangling doc links)


AirRocker:
## What was observed

Installing with `npx impeccable update` (CLI `impeccable@3.5.0`, skill `4.0.4`) produces a skill tree that is missing `reference/degraded/` — 4 files that the `skill-v4.0.4` tag and its release asset both ship:

```
.claude/skills/impeccable/reference/degraded/asset-producer.md
.claude/skills/impeccable/reference/degraded/documenter.md
.claude/skills/impeccable/reference/degraded/finish-reviewer.md
.claude/skills/impeccable/reference/degraded/manual-edit-applier.md
```

The CLI installs from `https://impeccable.style/api/download/bundle/universal` (`cli/bin/commands/skills.mjs:566`, `API_BASE` at :23). That served bundle differs from the `universal.zip` asset attached to the `skill-v4.0.4` release, even though both declare `version: 4.0.4`:

| | served `/api/download/bundle/universal` | `skill-v4.0.4` release asset |
|---|---|---|
| size | 13,408,195 bytes | 14,654,654 bytes |
| entries matching `degraded` | **0** | **75** |
| `.claude/skills/impeccable/` entries | 163 | 170 |
| `argument-hint` | `"[shape · audit\|critique · …` | `"[craft\|shape · audit\|critique · …` |

Reproducible from a clean fetch, no local state required:

```sh
curl -sL https://impeccable.style/api/download/bundle/universal -o served.zip
gh release download skill-v4.0.4 --repo pbakaus/impeccable --pattern universal.zip

unzip -l served.zip      | grep -c degraded   # 0
unzip -l universal.zip   | grep -c degraded   # 75
```

## Consequence

The served bundle is internally inconsistent: it ships documents that link to files the same bundle omits. Three dangling links, in the installed `.claude` copy:

- `reference/new-work.md:105` → `[degraded/finish-reviewer.md](degraded/finish-reviewer.md)`
- `reference/new-work.md:107` → `[degraded/documenter.md](degraded/documenter.md)`
- `reference/visualize.md:45` → `[degraded/asset-producer.md](degraded/asset-producer.md)`

An agent following those links from an installed skill hits nothing. These are the no-subagent fallback path, so the failure is quiet — it only surfaces in the degraded case, which is exactly when the fallback is supposed to carry the run.

## Evidence this is packaging, not a local problem

[...]

PR #487, opened 2026-08-03
Drop deprecated craft alias from the generated argument-hint

abdulwahabone:
Part 1 of the fix for #475 (merge this first).

`craft` is a deprecated compatibility alias, but its `SKILL_CATEGORIES` entry kept it advertised in every generated `argument-hint`. This unmaps it: the hint stops listing craft, the alias keeps routing via the router table and `command-metadata.json`.

Needed before the impeccable-site fix for #475 deploys: the site will start building bundles with the current transformers, and without this line removed the served hint would re-advertise craft.

Validated with `bun run build` (hint now reads `[shape · audit|critique · ...]`) and `bun run test` (all pass). No version bump, no generated output staged.

AI-assisted change, prepared under maintainer direction.

### DIFF
--- a/scripts/lib/skill-categories.js
+++ b/scripts/lib/skill-categories.js
@@ -6,8 +6,10 @@
 
 export const SKILL_CATEGORIES = {
   // CREATE - build something new
+  // craft is deliberately unmapped: it is a deprecated compatibility alias
+  // (see the router table), and a command without a category stays out of
+  // the generated argument-hint while continuing to route.
   impeccable: 'create',
-  craft: 'create',
   shape: 'create',
   // EVALUATE - review and assess
   critique: 'evaluate',


scripts/lib/skill-categories.js
```diff
@@ -6,8 +6,10 @@
 
 export const SKILL_CATEGORIES = {
   // CREATE - build something new
+  // craft is deliberately unmapped: it is a deprecated compatibility alias
+  // (see the router table), and a command without a category stays out of
```
zccdyp4d82-cyber, line 10: يفكر فيا
الآن

zccdyp4d82-cyber: هلا

zccdyp4d82-cyber: سحر

zccdyp4d82-cyber: اذيال

PR #493, opened 2026-08-03
[security] Contain project-supplied hook.auditLog to the project root

ingnicolaboccato-lab:
`writeAuditLog` takes its target from `IMPECCABLE_HOOK_LOG` or, failing that, from `hook.auditLog` in `<project>/.impeccable/config.json`. It then expands `~/`, accepts absolute paths, and does `mkdirSync(recursive)` + `appendFileSync` with no containment check.

Those two sources are not equally trusted:

- `IMPECCABLE_HOOK_LOG` is set by the person running the harness.
- `hook.auditLog` ships with whatever repository you open — and this hook runs on **every** `Edit`/`Write` and on `Stop`.

So a checkout you cloned just to look at can have the hook create directories and append NDJSON anywhere the user can write. No exotic trick required: it is a plain string in a config file the repo carries.

### What this changes

Paths that came from the project config must resolve inside the project root. `IMPECCABLE_HOOK_LOG` keeps working anywhere, which is the point of having an env escape hatch.

The containment gate already existed and is used for reads — `isScanTargetInsideProject`, which canonicalizes symlinks before comparing. It simply was not applied to this write path. Reusing it keeps one definition of "inside the project" instead of adding a second, weaker one.

### Verification

Sandboxed harness with `HOME`/`USERPROFILE` redirected to a temp dir, five cases:

| Case | Expected | Result |
|---|---|---|
| project config, absolute path outside | rejected, no file created | pass |
| project config, `~/…` | rejected, nothing written to home | pass |
| project config, relative path inside project | still works | pass |
| project config, `../` traversal | rejected | pass |
| `IMPECCABLE_HOOK_LOG`, absolute path | still permitted | pass |

Patch is against `skill/scripts/`, so `./plugin` and the provider mirrors pick it up on the next build.

---

> [!NOTE]
> **Low Risk**
> Small, targeted security fix in hook audit logging with no change to env-controlled logging behavior; regression risk is limited to projects that relied on out-of-project `hook.auditLog` paths.
> 
> **Overview**
> **Closes a path where untrusted repo config could make the design hook write audit NDJSON outside the project.**
> 
[...]

Issue #515, opened 2026-08-05
hook.auditLog from the opened project's config writes outside the project (reproduced)


ingnicolaboccato-lab:
`writeAuditLog` takes its destination from the opened project's own config and writes there with no containment. Since the hook runs on every `Edit`/`Write`, opening a repository that ships an `.impeccable/config.json` is enough to have the hook create directories and append lines anywhere the user can write.

Reproduced twice against current `main`.

### Reproduction

Fixture: a project with `package.json`, one HTML file, and this `.impeccable/config.json` — nothing else.

**1. Absolute path outside the project**

```json
{ "hook": { "auditLog": "<scratch>/FUORI-DAL-PROGETTO/fuga.jsonl" } }
```

One synthesized `PostToolUse` event through the real `hook.mjs`:

```
cartella fuori dal progetto creata : True
file di log scritto fuori          : True
{"ts":"2026-08-05T09:12:52.306Z","event":"PostToolUse","harness":"claude","cwd":"…"}
```

The directory did not exist beforehand.

**2. Home-relative path**

```json
{ "hook": { "auditLog": "~/.ssh/impeccable-era-qui.jsonl" } }
```

With `HOME`/`USERPROFILE` redirected to a scratch directory so the real one stays untouched:

```
<HOME>\.ssh\impeccable-era-qui.jsonl  (618 byte)
```

`.ssh` did not exist in that home. The hook created it.

### Code

`skill/scripts/hook-lib.mjs`, `writeAuditLog`:

```js
let target = env?.IMPECCABLE_HOOK_LOG;
if (!target || typeof target !== 'string') {
  try { target = readConfig(baseCwd).auditLog; } catch { target = null; }
}
…
if (target.startsWith('~/')) {
  expanded = path.join(process.env.HOME || process.env.USERPROFILE || '.', target.slice(2));
} else if (path.isAbsolute(target)) {
  expanded = target;
} else {
  expanded = path.resolve(baseCwd, target);
}
fs.mkdirSync(path.dirname(expanded), { recursive: true });
fs.appendFileSync(expanded, line);
```

Both branches that leave the project — `~/` and absolute — are handled explicitly, and `..` traversal works through the third. There is no check that the resolved path stays inside the project.

### Why the two sources are not equally trusted

`IMPECCABLE_HOOK_LOG` is set by the person running the harness: pointing it anywhere is the point of an escape hatch, and it should keep working.

[...]

PR #517, opened 2026-08-05
[security] Contain project-supplied hook.auditLog to the project root (#515)

ingnicolaboccato-lab:
Closes the defect reproduced in #515.

`writeAuditLog` takes its destination from `IMPECCABLE_HOOK_LOG` or, failing that, from `hook.auditLog` in `<project>/.impeccable/config.json`. It then expands `~/`, accepts absolute paths, and does `mkdirSync(recursive)` + `appendFileSync` with no containment check.

Those two sources are not equally trusted:

- `IMPECCABLE_HOOK_LOG` is set by whoever runs the harness.
- `hook.auditLog` ships with whatever repository gets opened, and this hook runs on **every** `Edit`/`Write` and on `Stop`.

So a checkout cloned just to look at can have the hook create directories and append NDJSON anywhere the user can write. #515 has it reproduced twice: an absolute path outside the project, and `~/.ssh/` with `HOME` redirected so the real one stayed untouched. In both cases the hook created the directory and wrote the file.

### The change

The containment gate already exists and already guards the read paths — `isScanTargetInsideProject`, which canonicalizes symlinks before comparing. It was simply never applied to this write. Reusing it keeps one definition of "inside the project" rather than adding a second, weaker one.

Paths that came from the project config must resolve inside the project root. `IMPECCABLE_HOOK_LOG` keeps working anywhere, which is the point of having an escape hatch.

### Tests

The automated review on #493 was right to withhold approval: one existing case asserted the behavior this prohibits. `resolves config auditLog from entry.cwd` put the destination in the **parent** of the project root. Its subject is which cwd the config is read from, not where the file may land, so the fixture moves the destination inside the project and still isolates that.

Four boundary cases added, none of which existed:

| Case | Expected |
|---|---|
| configured absolute path outside the project | refused, nothing created |
| configured `../` traversal | refused, nothing created |
| configured `~/.ssh/…` with `HOME` redirected | refused, `.ssh` not created |
| `IMPECCABLE_HOOK_LOG` outside the project | still written |

Checked that they can fail: against `ae5e9510` the three `refuses` fail and the other two pass. The reworked `entry.cwd` test passes on both — it was not weakened into something only the patched tree satisfies.

One caveat if you run `tests/hook.test.mjs` yourself: `IMPECCABLE_HOOK_DISABLED=1` in the environment makes the hook return nothing, and every `preToolUse` test then reads `allow` where it expects `deny`. Nine failures come from that alone. Unset it before reading the numbers.

### Process

Opening ahead of an explicit go-ahead, which I know is what closed #493 — though that one was closed for `no-linked-issue`, and the missing issue is now #515, where the request has been sitting since 2026-08-05. Submitting the finished work rather than a second promise of it; if the policy still calls for a close, close it and I will reopen on your word.

Happy for you to write it yourself instead. The reproduction is the part I care about.

---

> [!NOTE]
> **Medium Risk**
> Security fix on filesystem writes triggered on every Edit/Write/Stop; behavior change for projects that configured `hook.auditLog` outside the repo, while the env escape hatch remains.
> 
> **Overview**
> **Closes a path where untrusted repo config could make the hook append audit NDJSON outside the project** (e.g. absolute paths, `..`, or `~/…`), since `hook.auditLog` in `.impeccable/config.json` is not as trusted as operator-set `IMPECCABLE_HOOK_LOG`.
> 
> `writeAuditLog` now tracks when the destination came from project config and **refuses** to create directories or append if the resolved path is not inside the project root, reusing the same **`isScanTargetInsideProject`** canonical containment used for scan targets. **Env override `IMPECCABLE_HOOK_LOG` is unchanged** and may still point outside the project.
> 
> Tests cover refusal for outside absolute, `..` traversal, and home expansion, plus confirmation that env override still works; an existing `entry.cwd` fixture was adjusted so it no longer relied on writing outside the project.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27f9ed3454d7478b24d04aed46d0369e7aee6124. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skill/scripts/hook-lib.mjs
+++ b/skill/scripts/hook-lib.mjs
@@ -1483,9 +1483,19 @@ export function writeAuditLog(env, entry, cwd = process.cwd()) {
   // process cwd can differ from the project being edited.
   const baseCwd = entry && typeof entry.cwd === 'string' && entry.cwd ? entry.cwd : cwd;
   // Env wins; otherwise fall back to the unified config's hook.auditLog path.
+  // The two sources are NOT equally trusted: IMPECCABLE_HOOK_LOG is set by the
+  // person running the harness, while hook.auditLog is read from
+  // <project>/.impeccable/config.json — i.e. it ships with whatever repository
+  // you happen to open. Since this hook runs on every Edit/Write, an untrusted
+  // checkout could otherwise point it at an absolute or ~/ path and have the
+  // hook create directories and append lines anywhere the user can write.
   let target = env?.IMPECCABLE_HOOK_LOG;
+  let fromProjectConfig = false;
   if (!target || typeof target !== 'string') {
-    try { target = readConfig(baseCwd).auditLog; } catch { target = null; }
+    try {
+      target = readConfig(baseCwd).auditLog;
+      fromProjectConfig = true;
+    } catch { target = null; }
   }
   if (!target || typeof target !== 'string') return false;
   try {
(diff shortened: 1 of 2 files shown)


abdulwahabone:
Thanks for taking the time to contribute. Impeccable is issue-first for outside contributions. Although #515 is linked, I could not find explicit approval or a request from @pbakaus or @abdulwahabone to open a PR. Opening a PR before that approval is against our contribution policy. Please wait for explicit maintainer approval, then open a fresh PR. This policy check was performed automatically by Codex; if it missed existing approval, please point a maintainer to it.

PR #537, opened 2026-08-07
Resolve repo-root DESIGN.md through monorepo subpackage boundaries

willtmc:
## The bug

`findDesignRoot` in `cli/engine/design-system.mjs` walks up from a scan target looking for `DESIGN.md`, and stops at the first directory carrying a project marker (`.git` / `package.json` / `.impeccable`) that has no `DESIGN.md`. That boundary rule is what keeps sibling projects isolated — but in a monorepo it creates a blind spot: a subpackage's `package.json` is such a directory, so it hides the repo-root `DESIGN.md` from every CLI scan of files under it.

Real-world case that surfaced this: a Rails monorepo with `DESIGN.md` and `.git` at the repo root and an npm subpackage at `web/`. Any `impeccable detect` run against files under `web/` silently resolved *no* design system, so `DESIGN.md` violations (off-palette fonts, colors, radii) in the app's actual frontend went unreported.

## The fix

Keep `.git` and `.impeccable` as hard boundaries, but treat a `package.json` that is nested inside a git repository as a monorepo subpackage rather than an independent project — the walk continues upward. This can never leak across repos: the walk still stops at the `.git` entry itself (a directory in a normal checkout, a file in a git worktree — `existsSync` covers both).

Two small helpers implement this:

- `isProjectBoundary(dir)` — `.git`/`.impeccable` always bound; `package.json` bounds only when no ancestor is a git repo.
- `hasGitAncestor(startDir)` — walks up checking for a `.git` entry.

Behavior outside git repos is unchanged: a bare `package.json` still marks an independent project, so the sibling-isolation guarantees from the cross-project contamination fix are preserved.

## Tests

8 new tests, all passing alongside the existing suites (`node --test tests/design-system.test.mjs tests/detect-cli-design-contamination.test.mjs` → 28 pass, 0 fail):

**`tests/design-system.test.mjs`** (7 unit tests on `findDesignRoot` / `loadDesignSystemForTarget`):
- monorepo subpackage resolves the repo-root `DESIGN.md`
- same, with `.git` as a worktree *file* instead of a directory
- `loadDesignSystemForTarget` applies the repo-root system to a subpackage file
- nearest `DESIGN.md` wins (subpackage `DESIGN.md` beats repo root)
- `package.json` outside any git repo still bounds (sibling isolation)
- a nested `.git` repo never inherits the outer repo's `DESIGN.md`
- `.impeccable` stays a hard boundary even inside a git repo

**`tests/detect-cli-design-contamination.test.mjs`** (1 e2e): a full `detect` run applies the repo-root `DESIGN.md` to a file under the subpackage boundary.

The pre-existing contamination tests (cwd project A must not govern project B's files) still pass unmodified.

---

> [!NOTE]
> **Low Risk**
> Localized change to design-root path resolution with broad regression tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes **silent skips of `DESIGN.md` enforcement** when scanning files under a monorepo subpackage (e.g. `web/` with its own `package.json` but no local `DESIGN.md`). The upward walk used to treat any `package.json` as a hard project boundary and stop with no design system.
> 
> **`findDesignRoot`** now delegates boundary detection to **`isProjectBoundary`** and **`hasGitAncestor`**: `.git` and `.impeccable` still always stop the walk; `package.json` stops only when there is **no** git ancestor (standalone projects). Inside a repo, nested `package.json` is ignored so the walk can reach a repo-root `DESIGN.md` (still capped at `.git`, including worktree `.git` files). Nearest `DESIGN.md` still wins; nested repos and `.impeccable` boundaries behave as before.
> 
> The same logic is mirrored across **`cli/engine/design-system.mjs`** and every bundled **impeccable** detector copy. **Tests** cover unit resolution, `detect` e2e, and hook bundle parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8ca9082cedbb3188e4c1ee579fac77b54b3c7ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.agents/skills/impeccable/scripts/detector/design-system.mjs
+++ b/.agents/skills/impeccable/scripts/detector/design-system.mjs
@@ -558,6 +558,10 @@ function designSystemStartDir(targetPath, cwd = process.cwd()) {
 //   - A directory carrying a project marker (.git / package.json / .impeccable)
 //     but no DESIGN.md is a project BOUNDARY: the walk stops with no design
 //     system, so a sibling project never inherits a parent's or cwd's rules.
+//     Exception: a package.json nested inside a git repository marks a monorepo
+//     subpackage, not an independent project — the walk continues so a
+//     repo-root DESIGN.md still governs files under the subpackage. The walk
+//     can never escape the repo, because the .git directory itself stops it.
 //   - Reaching the home directory / filesystem root with neither means no
 //     design system at all — never process.cwd()'s.
 //
@@ -568,12 +572,35 @@ export function findDesignRoot(startDir) {
   const homeDir = path.resolve(os.homedir());
   while (true) {
     if (resolveDesignMdPath(dir)) return { dir, hasDesign: true };
-    if (PROJECT_ROOT_MARKERS.some((marker) => fs.existsSync(path.join(dir, marker)))) {
-      return { dir, hasDesign: false };
-    }
+    if (isProjectBoundary(dir)) return { dir, hasDesign: false };
     if (dir === homeDir) return null;
     const parent = path.dirname(dir);
     if (parent === dir) return null;
+    dir = parent;
+  }
+}
+
+// .git and .impeccable always mark a project root. package.json marks one only
+// outside a git repository: inside one it is a subpackage (e.g. web/ in a
+// monorepo) whose files are still governed by a design root higher in the same
+// repo.
+function isProjectBoundary(dir) {
+  if (PROJECT_ROOT_MARKERS.some(
+    (marker) => marker !== 'package.json' && fs.existsSync(path.join(dir, marker)),
+  )) {
+    return true;
+  }
+  return fs.existsSync(path.join(dir, 'package.json')) && !hasGitAncestor(path.dirname(dir));
+}
+
+// True when `startDir` or any ancestor contains a .git entry (a directory in a
+// normal checkout, a file in a git worktree — existsSync covers both).
+function hasGitAncestor(startDir) {
+  let dir = path.resolve(startDir);
+  while (true) {
+    if (fs.existsSync(path.join(dir, '.git'))) return true;
+    const parent = path.dirname(dir);
+    if (parent === dir) return false;
     dir = parent;
   }
 }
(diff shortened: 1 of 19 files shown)


willtmc:
Closed the accidental duplicate #538 (same engine change, opened in parallel). Its two extra end-to-end CLI tests — the worktree .git-file case and the no-inherit-from-cwd repo-boundary case — are now included here, so this PR carries the full test coverage from both.

willtmc:
Greptile's P1 is confirmed and now addressed in c8ca9082.

**What was wrong:** every generated detector bundle (`plugin/skills/impeccable/scripts/detector/design-system.mjs` plus the 14 provider trees like `.claude/`, `.cursor/`, `.agents/`) was still byte-identical to the pre-fix engine, so the packaged hook stopped at a subpackage `package.json` and reported its files clean instead of applying the repo-root `DESIGN.md`. The drift was invisible to CI because the "Verify generated tracked outputs" step runs `bun run build` (`--skip-root-sync`), which never regenerates those trees.

**What c8ca9082 does:**
- Resyncs all 15 `detector/design-system.mjs` copies from `cli/engine` via the full build, so the shipped detector carries `isProjectBoundary`/`hasGitAncestor`.
- Adds a test in `tests/hook-build.test.mjs` that imports each shipped detector bundle (`.claude`, `.cursor`, `.agents`, `plugin`) and asserts `findDesignRoot` resolves a repo-root `DESIGN.md` through a monorepo subpackage. Mutation-checked: it fails against the stale bundle, passes after resync.

**Deliberately left out:** the Live runtime staleness Greptile also mentions (`live-browser.js`, `browser-script-parts.mjs`, the new `live/ui-surfaces.mjs` copies) and the #520 color-parsing resync predate this branch — main's own generated trees (tip aee6ce93) have the same drift, so I left that to the `sync-generated-output` workflow rather than folding unrelated churn into this PR. Happy to include a full resync here instead if you'd prefer.

abdulwahabone:
Hey @willtmc, thanks for taking the time to put this together and test it thoroughly.

We follow an issue-first process for outside contributions. Changes need to be discussed in an issue and explicitly approved by @pbakaus or @abdulwahabone before a PR is opened. I couldn’t find that prior approval linked here.

Please open an issue first, or link the approved discussion if I missed it, and wait for a maintainer’s go-ahead before opening a PR.

Thanks for understanding!

_AI assistance was used to prepare and post this comment._

Issue #486, opened 2026-08-03
broken-image: false positives on `<img>` mentioned in code comments (wrapper components that take src as a prop)


ethantran12345:
## Summary

In the regex/text engine, `broken-image` fires on the literal string `<img>` written inside **code comments**. In a React/TypeScript codebase this is common: image *wrapper* components take `src` as a prop, so their JSDoc naturally says things like "Extra classes on the `<img>` itself" — and every one of those doc lines is reported as a shipped broken image.

On our repo (React 18 + TS, ~600 UI files) this was 5 of 13 non-advisory findings — 38% of the actionable signal, all false. Because the noise is concentrated on precisely the well-documented wrapper components, the rule is loudest where the code is best.

## Root cause

`scripts/detector/engines/regex/detect-text.mjs:238`

```js
{ id: 'broken-image', regex: /<img\b(?:(?!\bsrc\s*=)[^>])*>/gi, ... }
```

`<img>` in prose matches: `<img` + zero `[^>]` chars + `>`. The text engine does not strip comments before matching, so any `//`, `/* */`, or JSDoc mention of the tag is a finding.

## Repro

Three of our wrapper components, each a single JSDoc/comment line, no real defect:

| File | Line | Comment text that fires |
|---|---|---|
| `src/components/ui/thumb-image.tsx` | 14 | ` * <img> that serves a Supabase Storage transform (thumbnail) instead of the` |
| `src/components/ui/blur-up-image.tsx` | 44 | `  /** Extra classes on the <img> itself. */` |
| `src/lib/imageTransform.ts` | 11 | ` * Callers must keep the ORIGINAL url as an <img> onError fallback: if the` |

Two more from the same run, same shape — one prop doc, one explanatory block comment:

- `src/components/ImageGalleryCarousel.tsx:41` — `/** Extra classes applied to each slide <img> (e.g. object-top). */`
- `src/components/roommate/RoommateCard.tsx:157` — `... which is not a valid pattern on an <img>` (inside a `/* */` block)

Minimal case:

```ts
/** Extra classes on the <img> itself. */
export interface Props { imgClassName?: string }
```

```
$ npx impeccable detect repro.ts
broken-image  repro.ts:1  <img>
```

Note `src/lib/imageTransform.ts` contains **no JSX at all** — it is a pure URL-building module. A `.ts` file with zero markup should never be able to trip an `<img>` rule.

[...]

PR #549, opened 2026-08-09
Fix: stop reading comments as markup in the regex engine

nathannascimentofivix:
## The problem

The line matchers in `cli/engine/engines/regex/detect-text.mjs` scan raw source, so a comment that merely **names** a tag is scanned as if it were that tag.

A block comment explaining why an image is proxied:

```jsx
/* The CDN answers 403 for an `<img>` loaded from another origin, so the
   picture goes through our own route. */
```

reports `broken-image` in a file that contains no `img` element at all. Every matcher has the same exposure — documentation about markup gets read as markup. In one 256-file Next.js app this produced 11 phantom `broken-image` findings and 1 phantom `gray-on-color` (a docblock quoting `text-slate-700`).

## The fix

`maskComments()` blanks comment bodies before the line matchers run.

Bodies are replaced with **spaces** rather than removed, so byte offsets, line numbers and column math are unchanged — reported lines stay correct and line-scoped inline ignores keep working.

It is a small scanner rather than a regex on purpose. Stripping `//` with a regex would swallow the rest of `"https://x"` and could **hide a real finding**, which is worse than the false positive being fixed. The scan tracks strings and template literals and only treats a comment opener as a comment when it opens outside one.

## Validation

**New tests** — `node --test tests/detect-comments-are-not-code.test.mjs`, 7 passing. They pin **both directions**, because a fix that only silenced the rule would pass the first half:

- block / line / HTML comment naming an img tag → not flagged
- `` with no `src` → still flagged
- `` → still flagged
- `` on the line after a `"https://…"` string → still flagged
- reported line number still points at the real line

**No regression** — `node --test` on `detect-antipatterns-fixtures`, `detect-antipatterns-browser` and `framework-fixtures` gives **identical pass/fail counts with and without this change** (41/32, 0/27, 211/5). Those failures are this environment lacking `bun` and build artifacts, not this patch; I baselined by stashing the change and re-running.

**Measured impact** — on the 256-file app above: 14 findings became 2. All 12 that disappeared were comments, checked one by one. No code line stopped being seen.

## Notes

Generated provider output is intentionally omitted, per the Generated Provider Output Policy in `AGENTS.md`. I could not run `bun run build` / `build:browser` / `build:extension` — `bun` is not available in this environment — so the browser and extension bundles will need regenerating on your side.

---

> [!NOTE]
> **Low Risk**
> Scoped change to regex detection preprocessing with regression tests; no security, data, or API surface impact.
> 
> **Overview**
> **Regex line matchers** no longer treat comment prose as markup. Before `runRegexMatchers` runs on the full file, **`maskComments()`** replaces comment bodies with spaces (keeping newlines and offsets) so tags or class names mentioned only in `//`, `/* */`, or `<!-- -->` do not trigger rules like `broken-image` or `gray-on-color`.
> 
> The mask is a small scanner, not a naive regex strip: it skips strings and template literals so a `https://` URL inside quotes is not mistaken for a line comment and real findings on the same line are not hidden.
> 
> New tests in `tests/detect-comments-are-not-code.test.mjs` assert comments are ignored, real `` issues still fire, URL-in-string behavior is safe, and reported line numbers stay correct.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f930bfb30931ac134b89435e7a290f139d871aa3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cli/engine/engines/regex/detect-text.mjs
+++ b/cli/engine/engines/regex/detect-text.mjs
@@ -18,6 +18,71 @@ const hasBorderRadius = (line) => /border-radius/i.test(line);
 const isSafeElement = (line) => /<(?:blockquote|nav[\s>]|pre[\s>]|code[\s>]|a\s|input[\s>]|span[\s>])/i.test(line);
 
 
+/**
+ * Blank out comment bodies so the line matchers never read prose as markup.
+ *
+ * The line matchers scan raw source, so a comment that merely NAMES a tag is
+ * scanned as if it were that tag. A block comment explaining "the CDN answers
+ * 403 for an img loaded from another origin" reports `broken-image` in a file
+ * that contains no img element at all. Every matcher has the same exposure —
+ * documentation about markup gets read as markup.
+ *
+ * Comment bodies are replaced with spaces instead of removed, so byte offsets,
+ * line numbers and column math are unchanged. Reported lines stay correct and
+ * line-scoped inline ignores keep working.
+ *
+ * This is a small scanner rather than a regex on purpose: `"https://x"` has to
+ * survive. Stripping `//` with a regex would swallow the rest of that string
+ * and could HIDE a real finding, which is worse than the false positive being
+ * fixed. So the scan tracks strings and template literals, and only treats a
+ * line or block comment as a comment when it opens outside one.
+ */
+function maskComments(source) {
+  const out = source.split('');
+  const blank = (from, to) => {
+    for (let k = from; k < to && k < out.length; k++) if (out[k] !== '\n') out[k] = ' ';
+  };
+  let i = 0;
+  const n = source.length;
+  while (i < n) {
+    const c = source[i];
+    const d = source[i + 1];
+    if (c === '"' || c === "'" || c === '`') {
+      const quote = c;
+      i++;
+      while (i < n) {
+        if (source[i] === '\\') { i += 2; continue; }
+        if (source[i] === quote) { i++; break; }
+        i++;
+      }
+      continue;
+    }
+    if (c === '/' && d === '/') {
+      let j = i;
+      while (j < n && source[j] !== '\n') j++;
+      blank(i, j);
+      i = j;
+      continue;
+    }
+    if (c === '/' && d === '*') {
+      const end = source.indexOf('*/', i + 2);
+      const j = end === -1 ? n : end + 2;
+      blank(i, j);
+      i = j;
+      continue;
+    }
+    if (c === '<' && source.startsWith('<!--', i)) {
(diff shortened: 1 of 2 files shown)


pbakaus:
Thanks for the careful write-up and for testing both directions; the "a fix that only silenced the rule would pass the first half" framing is exactly right, and the line-number preservation test is a nice touch.

Unfortunately the core of this change duplicates and then regresses a fix that already landed. Issue #486 reported this same false positive, and PR #490 (commit de7b7284, merged Aug 3) added `stripJsComments()` in `detect-text.mjs`. It is a position-preserving scanner that understands regex literals, JSX, template literals, and `${}` expressions, and the full-file matcher pass already runs on its output. Your branch is based on that main, and 6 of your 7 new tests pass on main without this patch. I suspect your 256-file measurement was taken with an older published CLI that predates #490.

The new `maskComments()` replaces that stripped input with a scan of raw content, and because it only understands quoted strings, it treats any bare `//` as a comment. Three concrete regressions, verified against `detectText`:

1. `const re = /\/\//; const el = ;` no longer reports broken-image (the regex literal's trailing slashes read as a line comment). Main reports it.
2. `const x = <p>Docs at https://example.com</p>; const el = ;` no longer reports broken-image (bare URL in JSX text). Main reports it.
3. In a `.css` file, `background: url(https://cdn.example.com/x.png); color: #999;` has everything after `https:` blanked, even though plain CSS has no `//` comment syntax at all, which corrupts the block-context pass.

Hiding a live finding is worse than the false positive, as your own PR description argues, so this cannot merge as is. I'm closing this PR.

There is one genuinely new thing here: HTML comments in `.html`-like files are not covered by `stripJsComments` (it only runs for JS extensions), and your HTML-comment test really does fail on main. That is worth fixing. The shape I would take: a narrow, position-preserving mask for `<!-- -->` only, applied to non-JS extensions inside the existing strip pipeline next to `stripCssInJsComments`, leaving the JS path exactly as #490 built it, plus tests that keep your both-directions structure and add the unquoted-URL and regex-literal shapes above as still-flagged cases. If you'd like to take that on, please open an issue describing the gap first and wait for a maintainer go-ahead, per the contributing policy.

One process note: this repo is issue-first for outside contributions (see the Contributing section in AGENTS.md). This PR does not reference an issue and was not discussed with a maintainer beforehand. The repo also requires AI assistance to be disclosed in commits and PR text; if any of this was AI-assisted, please say so in future contributions.

*This analysis and reply were prepared with AI assistance (Claude Code); the regression reproductions were verified by running `detectText` on both branches.*

Issue #512, opened 2026-08-05
[Bug] `hooks reset` re-arms a previously disabled hook: it leaves the manifest behind while `DEFAULT_CONFIG.enabled` is `true`


khanxxy:
## Summary

`hooks off` disables the hook correctly. The problem is the next step: `hooks reset` deletes the config, the consent record and the cache, but never removes the hook entries it wrote into the harness manifests. Because `DEFAULT_CONFIG.enabled` is `true`, the missing config falls back to "on" — and the surviving manifest entry invokes the hook again.

So the revocation does not survive a reset. A user who deliberately turned the hook off and then cleaned up ends up with it armed.

Version: 4.0.4 (tag `skill-v4.0.4`, commit `9a949fb543d44cfb406f61bcab99d95d7f12cf1d`). Also present on current `main` (`ae5e95101a6979e7f7973a4ff57680b3c7adc1ec`) — line numbers differ, both are given below.

## Reproduction

Fixture with the skill linked into `.agents/skills/impeccable`:

```text
fresh         hooks status  →  enabled     (no manifest, no consent record)
hooks on      →  consent record + manifests written
hooks off     →  disabled                  ← correct
hooks reset   →  both config files deleted, manifest entries remain
hooks status  →  enabled                   ← the hook is armed again
```

The first line is a second, milder symptom: a fresh checkout with no consent record at all already reports `enabled`.

## Cause

`skill/scripts/hook-lib.mjs`, `DEFAULT_CONFIG` — line 151 on both the tag and `main`:

```js
export const DEFAULT_CONFIG = Object.freeze({
  enabled: true,
```

`skill/scripts/hook-admin.mjs`, `reset(cwd)` — line 682 at the tag, line 742 on `main`. It removes:

- the config file and the local config file, stripping the `hook` / `detector` subtrees,
- the cache file and the pending file.

It never calls `pruneImpeccableHookFromManifest()` — defined at line 484 at the tag, 512 on `main` — which is currently only reached from `repairHookManifests()` (the call sites are line 373 at the tag, 401 on `main`). The manifests written by `hooks on` therefore outlive `hooks reset`, and the default supplies the `enabled` they need.

## Impact

There is no error and no hint in the output — `hooks status` simply reports `enabled` again. For a mechanism whose purpose is explicit consent, a reset that re-arms is the one failure direction that should not be possible.

## Suggested fix

[...]

PR #551, opened 2026-08-10
fix: hooks reset no longer re-arms a disabled hook

SomSamantray:
## Before opening

This repo is issue-first for outside contributions. I am not `pbakaus` or `abdulwahabone`, and there is no maintainer approval or request on the linked issue — opening this in good faith since the fix matches the reporter's own suggested fix exactly, with the understanding it may be closed without review per repo policy.

- Linked issue: #512
- Contributor status:
  - [ ] I am `pbakaus` or `abdulwahabone`
  - [ ] A maintainer approved this PR in the linked issue

## What was broken

`hooks off` correctly disables the hook. Running `hooks reset` afterward silently re-enabled it: `reset()` deleted the config file (including the `enabled: false` the user set) but never touched the hook manifests it wrote into `.claude/settings.local.json`, `.codex/hooks.json`, `.cursor/hooks.json`, or `.github/hooks/impeccable.json`. Because `DEFAULT_CONFIG.enabled` defaulted to `true`, the next config read treated the missing key as "on," and the surviving manifest entries fired again.

Repro from the issue:
```
fresh         hooks status  →  enabled     (no manifest, no consent record)
hooks on      →  consent record + manifests written
hooks off     →  disabled                  ← correct
hooks reset   →  both config files deleted, manifest entries remain
hooks status  →  enabled                   ← the hook is armed again
```

## The fix

Exactly the reporter's own two-part suggestion:

1. **`DEFAULT_CONFIG.enabled: false`** (`skill/scripts/hook-lib.mjs`). Absence of configuration now means "not consented" — only `hooks on` writes `true`, together with the consent record and the manifest entries.
2. **`reset()` prunes impeccable's entries from every installed manifest** (`skill/scripts/hook-admin.mjs`), reusing the existing `pruneImpeccableHookFromManifest()` — the same function `repairHookManifests()` already calls, so no new stripping logic was written. Unlike the install path, the prune loop is **not** gated on the skill folder still existing, so a reset mid-uninstall (skill files already removed, manifest cleanup still pending) still works.

One deliberate scope narrowing versus a first pass at this fix: `reset()` only prunes `destRel` (e.g. `.claude/settings.local.json`), never `sharedDestRel` (e.g. a team-committed `.claude/settings.json`) — matching the write-scope asymmetry `hooks on`/`repairHookManifests()` already maintain (they only ever *read* the shared file to check existing coverage, never write it). A single developer's local `reset` should not strip the hook from a file the whole team shares.

With both changes, the issue's exact sequence now ends disabled:
```
fresh → disabled, on → installs into every configured harness, off → disabled,
reset → removes config, consent, and manifests, final status → disabled
```

## Testing

- New tests in `tests/hook.test.mjs`: `reset()` happy path, all-four-providers, no-manifests-installed (behavior unchanged), no-marker (manifest untouched), skill-folder-absent (mid-uninstall), shared-manifest-untouched, and the full `on`/`off`/`reset`/`status` sequence reproducing the issue verbatim.
- `DEFAULT_CONFIG.enabled: false` has a wider blast radius than the line count suggests: roughly a dozen `runHook()`/`runStopHook()` test fixtures relied on the old `true` default to reach real detection logic on a bare temp dir with no config file. Added a `mkEnabledTmp()` helper and swapped it into every affected fixture; one test whose premise (a clean edit with literally zero `.impeccable/` footprint) became impossible to construct once enabling requires a config file was removed as redundant with the adjacent "opted-in project" test that already covers the same case correctly.
- Full suite: 214/223 passing on this Windows dev environment, matching the same 9 pre-existing failures on `main` (Windows symlink-permission and path-separator issues unrelated to this change) — zero regressions introduced.
- Reviewed with `ce-code-review` (correctness, testing, adversarial, project-standards). Two findings applied: the shared-manifest write-scope fix above, and a latent (currently inert) default mismatch in `mergeHookConfig()`. Two lower-severity findings — a pre-existing substring-matching quirk in the manifest-stripping logic, and no file locking against concurrent `hooks` invocations — are pre-existing/out-of-scope and recorded in `docs/residual-review-findings/441bf5b0.md` rather than actioned here.

AI-assisted change (Claude Code), following the reporter's own repro and suggested fix in #512.

---

> [!NOTE]
> **Medium Risk**
> Changes default hook behavior and reset/manifest semantics across CLI, context, and hook runtime; mistakes could leave hooks silently off or partially reset, but behavior is heavily tested.
> 
> **Overview**
> Fixes **#512**: a `hooks reset` after `hooks off` could leave manifest entries in place while config was gone, and missing `hook.enabled` was treated as **on**.
> 
> **Opt-in by default:** `DEFAULT_CONFIG.enabled` is now **false** in `hook-lib.mjs`, and `hookEnabledAt()` in `context.mjs` matches that. A manifest alone no longer counts as an active hook; `MANUAL_DETECTOR_REQUIRED` and the runtime agree on when the hook is off.
> 
> **`hooks reset`:** After clearing hook/detector config (with **backup/rollback** if shared vs local config fails partway), reset **prunes Impeccable entries from local provider manifests only** (`destRel`, not team-shared `settings.json`). Config or manifest prune failures **throw** instead of reporting success with a stale armed state.
> 
> **Skills CLI install:** When install/update would wire hooks, it now writes **`hook.enabled: true`** to shared `config.json` via new `getHookEnabled` / `setHookEnabled`, but **does not** flip `enabled` back on if the user already set **`hooks off`** (`enabled: false`).
> 
> **Tests:** Broad hook test updates use `mkEnabledTmp()` where detection logic is under test; new coverage for reset pruning, rollback, and skills install enable behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ea5c01cdb3a736212015d240a43f0d360903f3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SomSamantray:
Both automated reviews caught real gaps — thanks. Fixed in `ee65cede`:

**Cursor Bugbot** — correct: `npx impeccable skills install` (the CLI's own install path, separate from `/impeccable hooks on`) recorded consent and wrote manifests but never wrote `hook.enabled: true`, and `context.mjs`'s `hookEnabledAt()` had its own independent hardcoded `true` default. Both would have silently broken under the new default. Fixed: `decideHookInstall()` now affirms `hook.enabled: true` via a new `setHookEnabled()` on every path that decides the hook should be on (including pre-existing consent recorded before this fix); `hookEnabledAt()`'s default now matches `hook-lib.mjs`'s.

**Greptile** — also correct: `reset()`'s pre-existing config-removal loop already swallowed fs errors silently, and my new manifest-pruning ran regardless — so a write failure (permissions, disk full) could leave `hook.enabled: true` on disk while the manifest is already gone, with `status` reporting enabled even though nothing invokes the hook anymore. `reset()` now surfaces that failure and aborts before touching any manifest.

Full suite re-verified after both fixes: zero regressions (identical pre-existing Windows HOME/symlink environment failures on `tests/skills-cli.test.js`, verified against unmodified `origin/main` for comparison).

SomSamantray:
Fixed in \`c207e974\` — Greptile's finding was real: my previous fix threw *after* the config-removal loop, with no rollback, so a shared-config-succeeds-then-local-config-fails sequence left the shared config reset while the local one (which may still carry \`hook.enabled: true\`) was untouched — a partial state that's neither the old one nor the new one.

\`reset()\` now backs up each config file's exact bytes before mutating it and rolls back every already-applied mutation if a later step in the same call fails, so the config pair is genuinely all-or-nothing.

While in there, audited the rest of the function for the same class of gap rather than patching just this one report: \`pruneImpeccableHookFromManifest()\`'s writes weren't try/caught in the new pruning loop (one provider failing would crash uncaught after config was already reset, with no report of what succeeded), and the cache/pending removal loop silently swallowed failures inconsistently with everything else now surfacing them. Both fixed.

New test reproduces the exact two-file scenario Greptile's repro exercised and asserts the shared config is restored to its exact prior content. Full suite re-verified, zero regressions.

SomSamantray:
Fixed in \`8eb4b8c0\` — this was a real, still-open finding from an earlier round I'd missed replying to: \`decideHookInstall()\`'s \`enable()\` helper force-wrote \`hook.enabled: true\` on every affirmative path, including when consent was already \`accepted\`. Since consent and enabled are tracked on separate paths (consent by this CLI install flow, \`enabled\` by \`/impeccable hooks on|off\`), a user who accepted the hook long ago and later ran \`hooks off\` would have that opt-out silently re-armed by a routine \`skills install\`/\`update\`.

New \`getHookEnabled()\` reads the currently effective value (mirroring the runtime's own resolution order); \`enable()\` now only writes \`true\` when it isn't already explicitly \`false\`. Manifests still get wired up either way — only the enabled flag itself is now guarded. Two new tests cover both the guarded case and confirm a fresh install is unaffected.

I did a full audit of the review threads on this PR and don't see any other open findings — flag if I've missed something.

SomSamantray:
Fixed in \`45547f0b\` — fair catch: manifest-pruning failures were caught and reported in the returned message, but \`reset()\` still exited 0. A surviving manifest entry after a permissions/disk failure is the exact artifact this whole fix exists to prune, so burying that in ordinary-looking output gave no signal the reset was incomplete.

Manifest failures are now fatal too, same as config failures already were — config still ends up fully reset either way (fail-safe), but the command now correctly reports failure when manifest cleanup doesn't complete. New test proves both halves at once.

pbakaus:
Please do not submit PRs without explicit approval in the issue thread. That being said, I'm willing to make a one-time exception, but please remove the docs files cruft from the PR, and ensure that CI is not failing.

SomSamantray:
Done in 2ea5c01c:

- Removed `docs/plans/` and `docs/residual-review-findings/` — dropped both, kept only the code fix.
- Fixed CI: the symlinked-cwd `runHook()` test wrote its project files into a subdirectory of the test's `mkEnabledTmp()` cwd, so the enabled config that helper writes never covered the actual project root being scanned. Under the new disabled-by-default default, that made the test see the hook as disabled. Test now writes its own explicit `hook.enabled: true` config into the real project root. `npm run test:core` is green (674/674).

abdulwahabone:
Maintainer note: we independently reproduced this on current `main` (`0c251788`), not on this branch.

The writeup is #667. Short version: `hooks off` is the only persistent disable, `hooks reset` deletes that flag and leaves all four provider manifests in place, and a missing config still means on. After `on` → `off` → `reset`, `hooks status` printed `enabled`, the leftover Cursor/Claude/Codex/Copilot manifests still contained Impeccable commands, and a real `hook.mjs` PostToolUse event scanned `index.html` and emitted `gradient-text`.

#512 was the original report (code and fixture). #667 adds the runtime evidence: admin CLI output, leftover files, `runHook()` skip vs scan, and the spawned `hook.mjs` process.

This comment is not a review of the patch.

*Drafted with AI assistance, reviewed and posted by a maintainer.*

abdulwahabone:
I have created a leaner fix fir this issue in https://github.com/pbakaus/impeccable/pull/668
Thanks to @SomSamantray for flagging this issue and contributing the original fix.

PR #591, opened 2026-08-14
Fix uncaught spacing compression on comp-led builds

abdulwahabone:
## The Bug

Same run as #590 (cursor-control-8), different failure: the build packs the comp's content into roughly **half its relative height**. In the comp, a board row is about three of its own text-heights tall and the ticket cells carry real air; the build ships rows hugging their text and cells with the label crowding the value. Nothing in the flow records or checks spacing.

Research on screenshot-to-code says this is systematic, not a lapse: models exhibit pattern-completion bias — they perceive the comp's spacing correctly and still emit the genre's stock density. That drift runs both ways: an airy comp ships cramped (this run), and a dense comp ships loose. Spacing is documented as one of the least reliable properties in image-to-code, so it needs the same treatment color got in #590: turn perception into a recorded number once, then check against the number.

| Approved comp (rows ~3x their type size) | Shipped build (rows hugging text) |
| --- | --- |
| Approved comp | Shipped build |

## The Fix

Three text edits, stacked on #590 (they extend the same record and checkpoints):

- `skill/reference/visualize.md` — the brief record also carries the comp's **air**, read as ratios: key elements' heights and paddings as rough multiples of their own type size, section gaps and gutters as fractions of the comp's width. Ratios are the durable unit (comp and build widths differ), and estimation is the executable instruction; demanding pixel measurement from a JPEG invites fabricated numbers.
- `skill/reference/new-work.md` — air is a quantity in the comp-is-king comparison: the build's numbers come from the browser that captured the screenshots, where element boxes are exact. Tolerance is generous; the failure mode is wholesale (half or double the room), not small drift, and it binds in either direction.
- `skill/agents/impeccable-finish-reviewer.md` — **AIR** joins TYPE, MATERIAL, and GROUND as a mandatory fidelity-matrix row, anchored to the hero-repro capture the reviewer already verifies: it is taken at the comp's own dimensions, so equal air is directly visible at equal size. Tighter or looser than the comp is contradicted either way, mirroring GROUND's warmer-or-cooler form. The verdict pass judges fix-round regressions by the same matrix rules, so drift introduced by a fix batch is covered too.

## Validation

`bun run build` passes (prose gates clean). Generated provider output intentionally omitted.

AI-assisted change (Cursor), prepared under maintainer direction.


pbakaus:
This feels a tad specific...here too, I think we can't simply add more and more prose...feels like that's the wrong rabbit hole.

abdulwahabone:
> This feels a tad specific...here too, I think we can't simply add more and more prose...feels like that's the wrong rabbit hole.

I tried to not be specific here but you are right, its a wrong rabbit hole to go into. Closing it

PR #595, opened 2026-08-14
Fix: compile closest() selectors once per document in the static scanner (#575)

### LINKED ISSUE #575: Duplicate work in the static scanner dominates scan time
## Problem

`closest(selector)` walks from an element up to the root, testing each ancestor against a CSS
selector. The static-HTML version (`cli/engine/engines/static-html/css-cascade.mjs:837`) calls
`css-select`'s `is(node, selector)` at each step, and `is()` re-parses and re-compiles the
selector string on every call. An element N levels deep compiles the same selector N times.

Detection rules call `closest()` heavily. Profiling a scan of `tests/fixtures/antipatterns`
puts most of the run under `closest()`, and nearly all of that under selector compilation
rather than under the matching those selectors exist for. Selector parsing is the single
largest cost in the profile.

## Proposed solution

Compile each selector once per document and reuse it. `is()` already accepts a compiled matcher
in place of a string, so nothing else changes.

1. `detect-html.mjs`: add `compile: cssSelect.compile` to the module bag given to `StaticDocument`.
2. `StaticDocument`: hold a `Map` from selector string to compiled matcher.
3. `closest()`: look the selector up in that map.

Every call site passes a string literal or a module constant, so the map stays small and
bounded, and it lives on the document, so nothing survives a file. Store a failed compile as a
function that rethrows, so a bad selector still reaches the caller's `catch`.

`querySelectorAll` / `querySelector` recompile the same way, but they barely register in the
profile, so this stays scoped to `closest()`. After the change, `StaticDocument.is` has no
remaining caller.

A prototype cuts the fixture corpus and a single large page to roughly 40% of their current
time, with identical findings.

## Verifying

Output must not change:

- same findings across `tests/fixtures/antipatterns`
- `bun run test` passes with unchanged counts
- `scripts/benchmark-detector.mjs` for before/after numbers

## Notes

Analysis was done via Claude Opus 5 max effort.
I would be willing to work on this.

_Prepared with AI assistance (Claude)._

abdulwahabone:
The static scanner's `closest()` recompiled its selector string on every ancestor step via css-select's `is()`; `StaticDocument` now compiles each selector once per document and reuses the cached matcher (failed compiles cached as rethrowers, so bad selectors still return null). Fixes #575.

Findings are byte-identical before/after (394 across `tests/fixtures/antipatterns`, zero mismatches across all 105 `scripts/benchmark-detector.mjs` cases); fixture directory scan drops 239ms -> 149ms, and a deep-nested scratch corpus drops ~3.8s -> ~0.55s. `bun run test` green, plus two new regression tests that fail on the pre-fix engine. Generated provider output intentionally untouched.

Prepared with AI assistance (Cursor agent), directed by maintainer @abdulwahabone.

### DIFF
--- a/cli/engine/engines/static-html/css-cascade.mjs
+++ b/cli/engine/engines/static-html/css-cascade.mjs
@@ -835,10 +835,11 @@ class StaticElement {
     }
   }
   closest(selector) {
+    const matcher = this._doc.matcherFor(selector);
     let cur = this.node;
     while (cur && cur.type === 'tag') {
       try {
-        if (this._doc.is(cur, selector)) return this._doc.wrap(cur);
+        if (matcher(cur)) return this._doc.wrap(cur);
       } catch {
         return null;
       }
@@ -862,9 +863,10 @@ class StaticDocument {
     this.root = root;
     this.selectAll = modules.selectAll;
     this.selectOne = modules.selectOne;
-    this.is = modules.is;
+    this.compile = modules.compile;
     this.domutils = modules.domutils;
     this._wrappers = new WeakMap();
+    this._compiledSelectors = new Map();
     this._styleMap = new WeakMap();
     this._hoverStyleMap = new WeakMap();
     this._accentDashPseudo = new WeakSet();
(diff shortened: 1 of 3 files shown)


zieka (approved):
Should there be validation of modules.compile in the StaticDocument constructor in css-cascade to prevent silent fail?

abdulwahabone:
> Should there be validation of modules.compile in the StaticDocument constructor in css-cascade to prevent silent fail?

Thought about it, but I don't think it's needed. The only place that ever constructs StaticDocument is `detect-html.mjs, `and it builds the bag right there from the real imports.

If css-select fails to load we bail to the regex fallback before this class even exists. 

Old code had the same thing, a bag without is would silently null out closest() too. And if compile ever did go missing, the fixture suite would light up red since every closest() call would stop matching

pbakaus (approved):
LGTM

PR #606, opened 2026-08-17
Simplify provider hook smoke orchestration

pbakaus:
## Summary

- centralize Claude, Codex, and Cursor fixture paths, installed hook/admin paths, and hook-event construction in one provider contract
- route both smoke flows through one provider agent launcher while retaining provider-specific evidence and failure checks
- reuse the repository's shared CLI argument parser and add characterization tests for the public help/missing-repo contract

## Hindsight architecture review

**Hindsight is 20/20: if we were implementing these requirements today, would we design this file this way?** No.

The script encoded the same durable provider policy across five dispatch areas: fixture selection, admin-script selection, installed-hook invocation, agent-choice execution, and provider smoke execution. The leanest behavior-preserving design is one declarative provider contract plus one CLI-launch boundary. Provider-specific assertions remain separate because their evidence and failure semantics genuinely differ.

## Before / after

- primary file: **1,044 → 1,000 lines** (-44)
- named functions: **55 → 53**
- `if (provider === ...)` branches: **17 → 5**
- agent CLI launch definitions: repeated in both exception and provider smoke flows → defined once per provider
- argument parsing: private duplicate → existing shared `scripts/lib/cli-args.mjs`
- public contract: unchanged package entry point, CLI usage text, fixture paths, provider commands, logs, findings, and exit behavior

Across the two-file PR, the diff is 155 additions / 151 deletions (net +4 after adding 48 lines of characterization and regression coverage). Production code still shrinks by 44 lines.

## Validation

- `node --check scripts/smoke-provider-hooks.mjs`
- `node --test tests/cli-args.test.mjs` — 21 passed
- `bun run test` — full core, detector/browser, live, framework, and plugin E2E suite passed
- `git diff --check`

The destructive/credentialed `smoke:hooks` workflow itself was not invoked: it intentionally installs/uninstalls provider tooling and calls external agent CLIs. The public CLI boundary is characterized, the command/event contracts were preserved directly, and the repository's complete deterministic suite is green.

## Generated output

Generated provider and root harness output was intentionally omitted. This change only touches the source smoke script and its focused test coverage.

## Remaining risk

Low. The residual risk is limited to the opt-in provider smoke runner's external CLI integration, which is not suitable for unattended local execution because of its install/uninstall and authentication side effects.

## Authorization and AI assistance

Prepared by Codex with AI assistance under maintainer pbakaus's standing scheduled architecture-refactor authorization. This PR is ready for review and is not authorized for automatic merge.

### DIFF
--- a/scripts/smoke-provider-hooks.mjs
+++ b/scripts/smoke-provider-hooks.mjs
@@ -12,6 +12,8 @@ import { homedir, tmpdir } from 'node:os';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
+import { parseArgs } from './lib/cli-args.mjs';
+
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const prRoot = resolve(__dirname, '..');
 const defaultBundle = join(prRoot, 'dist', 'universal.zip');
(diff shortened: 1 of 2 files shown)


pbakaus:
> Re: comment 5317355005

Fixed in cf8f295d. The finding correctly identified a regression from the shared parser's boolean bare-flag sentinel. The script now translates only that sentinel back to its legacy string form, preserving the previous error behavior instead of broadening this refactor into a CLI UX change. Added characterization coverage for value-less repo, bundle, and providers options; 21 focused tests and the full suite pass. AI-assisted response from Codex under pbakaus's scheduled-refactor authorization.

PR #644, opened 2026-08-23
Fix: rewrite Grok project hooks to the global skill path (#642)

abdulwahabone:
Global skill installs left Grok's project hook pointing at `.grok/skills/...`, which does not exist. The hook rewrite now treats Grok like Claude and Codex so the command points at `~/.grok/skills/...`.
Fixes #642
AI assistance: Cursor Grok 4.6 implemented this change.
Validated with `bun run test` and `bun run test:cli-remote-e2e`.

Made with [Cursor](https://cursor.com)

Issue #642, opened 2026-08-23
[Bug] Grok Build hook points to project-level `hook.mjs` file even when skill only installed globally


felixarntz:
## What happened?

Using `npx impeccable install`, I installed the Impeccable skill globally for a few coding harnesses (including Grok Build), and then also installed the hook at the project level.

For Grok Build, the generated hook JSON file points to local `.grok/skills/impeccable/scripts/hook.mjs` despite the global skill installation - so this will error. See generated file:

```json
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write|MultiEdit",
        "hooks": [
          {
            "type": "command",
            "command": "[ ! -f \".grok/skills/impeccable/scripts/hook.mjs\" ] || ! node -e \"process.exit(Math.min(parseInt(process.versions.node,10),22)===22?0:1)\" 2>/dev/null || node \".grok/skills/impeccable/scripts/hook.mjs\"",
            "timeout": 5,
            "statusMessage": "Checking UI changes"
          }
        ]
      }
    ],
    "Stop": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "[ ! -f \".grok/skills/impeccable/scripts/hook.mjs\" ] || ! node -e \"process.exit(Math.min(parseInt(process.versions.node,10),22)===22?0:1)\" 2>/dev/null || node \".grok/skills/impeccable/scripts/hook.mjs\"",
            "timeout": 30,
            "statusMessage": "Design deep pass"
          }
        ]
      }
    ]
  }
}
```

For the other harnesses, this is fine, pointing to the globally installed skill's `hook.mjs` file, as expected.

## Steps to reproduce

1. Run `npx impeccable install`
2. Select harnesses that include Grok Build
3. Choose global skill install
4. Choose to add the hooks to the current project folder
5. Inspect the project-level `.grok/hooks/impeccable.json`

## Expected behavior

The hook should point to the globally installed skill's `hook.mjs` file instead.

PR #648, opened 2026-08-24
Fix: parse Grok Build camelCase hook stdin (#646)

### LINKED ISSUE #646: Grok Build design hook classifies native stdin as GitHub Copilot and never scans
Grok Build 1.0.5 does invoke `.grok/hooks/impeccable.json`. The process still exits 0 with `skipped: "no-file-path"` because `resolveHarness()` treats camelCase `toolName` + `toolInput` as GitHub Copilot (`toolName` + `toolArgs`). `normalizeGitHubEvent()` then looks at `toolArgs`, Grok's path stays on `toolInput`, and the detector never sees a file.

Stop is a second miss: Grok sends `hookEventName: "stop"`. `hook.mjs` only routes `hook_event_name === "Stop"` to `runStopHook`, so the deep pass never runs. Session id and `stopHookActive` are also camelCase. Grok then fires an observe-only Stop with `reason: "shutdown"` after `end_turn`; that one should stay silent.

PostToolUse stdout is ignored by Grok anyway. Stop `hookSpecificOutput.additionalContext` is the channel that can reach the model. Live capture from 2026-08-24 is in the local postmortem `docs/postmortem-grok-design-hook-2026-08-24.md` and the throwaway repo `impeccable-hook-fire-test`. Control: the same `hook.mjs` plus the same CSS emits `gradient-text` when given a Claude snake_case envelope.

This is independent of #642 (Grok global hook path).

Prepared with AI assistance from a live Grok stdin dump.

abdulwahabone:
Fixes #646

## Summary
- Grok Build 1.0.5 sends camelCase hook stdin (`toolName` + `toolInput`, `hookEventName: "stop"`). `resolveHarness()` treated that as GitHub Copilot (`toolName` + `toolArgs`), then looked at `toolArgs` and skipped with `no-file-path`. Stop never reached `runStopHook`.
- Normalize the Grok envelope onto the internal Claude shape. Discriminate Grok vs Copilot by `toolInput` vs `toolArgs`. Route `hookEventName: "stop"` into the deep pass, skip Grok's observe-only `reason: "shutdown"` Stop, and do not remember per-edit findings on Grok so Stop can still surface them (Grok discards PostToolUse stdout).
- Independent of #642 / #644 (global Grok hook path).

## Test plan
- [x] `node --test tests/hook.test.mjs` (236 pass), including captured-shape Grok `search_replace` / `stop` / `shutdown` cases and a `hook.mjs` subprocess routing check
- [x] `bun run test` (default suite)
- [x] `bun run build` (README + skill prose validators)
- [x] Live Grok 1.0.5 probe in `impeccable-hook-fire-test`: CSS edit via `search_replace`, Stop `end_turn` injected `gradient-text` / `overused-font` as `stop_hook_feedback`, nested agent cited those rules


abdulwahabone:
Tested on Grok, it works

PR #682, opened 2026-08-31
Fix: critique snapshot close path (#660)

### LINKED ISSUE #660: Critique snapshots have no close path, so polish inherits a stale backlog
## Summary

A critique snapshot is written as an archive and then presented as a live to-do list, with no way to mark it answered.

`$impeccable polish` and the no-argument routing menu read the latest file under `.impeccable/critique/` as a fix backlog. Nothing deletes that file, sets a resolved flag, or checks its `timestamp` against the target. `.gitignore` already treats per-run snapshots as local artifacts (`ignore.md` is the exception), so the intended lifetime is "this session" — but the skill still treats whatever is on disk as current work.

## What it costs

In [erikpr1994/peakhealth#992](https://github.com/erikpr1994/peakhealth/issues/992) a 2026-08-08 critique of the routine builder (score 21/40, seven Priority Issues) stayed in the tree after every finding had been fixed on `main`. A later session spent a full sub-agent pass (~200k tokens) proving each finding stale before it could act.

`.gitignore` would have stopped that file being committed. It would not have stopped polish from inheriting a leftover local snapshot, and it does not stop a consumer who *does* commit snapshots (this one did, because the consumer gitignore had mocks/review but not critique).

## Cause

`critique.md` names the snapshot an "archive/backlog for future commands". `polish.md` says: if `latest` exits 0, incorporate the P0/P1 findings. `routing.md` says: a low score or non-zero `p0`/`p1` → offer polish. `critique-storage.mjs` can `write` / `latest` / `trend` and has no close path.

The snapshot therefore outlives the work it described. A later run cannot tell "still open" from "already fixed" without re-deriving every finding against the current code.

## Suggested fix

Close a snapshot by deleting it. That is the cheapest resolved state: `latest` already treats absence as "no backlog", gitignore already wants snapshots off disk, and a new `$impeccable critique` writes a new file.

A `close` subcommand on the helper so agents do not hand-`rm` a glob.

### `skill/scripts/critique-storage.mjs`

Header, add the command:

```
 *   node critique-storage.mjs close <slug>
```

Export:

```js
/** Delete the latest snapshot for `slug`. Returns the path removed, or null. */
export function closeSnapshot(slug, { cwd = process.cwd() } = {}) {
  const latest = readLatestSnapshot(slug, { cwd });
  if (!latest) return null;
  fs.unlinkSync(latest.path);
  return latest.path;
}
```

CLI, next to `latest`:

```js
    case 'close': {
      const slug = coerceSlug(args[0]);
      if (!slug) { process.stderr.write('usage: close <slug-or-target>\n'); process.exit(1); }
      const closed = closeSnapshot(slug);
      if (!closed) { process.exit(2); }
      process.stdout.write(`${closed}\n`);
      return;
    }
```

Update the usage line in `default` to include `close`.

### `skill/reference/critique.md`

Purpose, last sentence:

```
- The chat response is the primary deliverable; the snapshot is an archive/backlog for future commands.
+ The chat response is the primary deliverable; the snapshot is an archive of that run. Later commands may read it as a starting backlog only while its findings still describe the code.
```

New section after **Persist the Snapshot**, before **Ask the User**:

```
### Close the Snapshot

A snapshot is open while the file exists. When every Priority Issue is fixed, declined, or written to `.impeccable/critique/ignore.md`, close it:

```bash
node {{scripts_path}}/critique-storage.mjs close "<resolved target>"
```

Exit 0 prints the path removed. Exit 2 means none existed. A new `$impeccable critique` writes a new snapshot. Do not leave a closed snapshot on disk: polish and routing will read it as an open backlog.
```

### `skill/reference/polish.md`

After the `latest` command:

```
- Exit 0 returns the latest snapshot; incorporate relevant P0/P1 findings and name the snapshot read. Exit 2 means none exists. Perform an independent pass either way.
+ Exit 0 returns the latest snapshot. Treat it as a starting backlog only while its findings still describe the target; when the target's source has moved on since `timestamp`, re-run `$impeccable critique` rather than inheriting the file. Exit 2 means none exists. Perform an independent pass either way.
```

In **5. Verify and finish**, after the source-diff sentence:

```
+ When this pass clears every Priority Issue it took from a snapshot, close that snapshot:
+
+ ```bash
+ node {{scripts_path}}/critique-storage.mjs close "<resolved target>"
+ ```
```

### `skill/reference/routing.md`

```
- `critique.latest` with a low `score` or non-zero `p0` / `p1` → `polish` (it reads that snapshot as its backlog), or re-run `critique` if the snapshot looks stale.
+ `critique.latest` with a low `score` or non-zero `p0` / `p1` → `polish` (it reads that snapshot as its backlog) only while the snapshot still describes the target; re-run `critique` when the target has been committed since the snapshot's `timestamp`.
```

### Tests

In `tests/critique-storage.test.mjs`, next to the `latest` CLI case:

```js
  it('close subcommand deletes the latest snapshot and exits 0', () => {
    writeSnapshot({ slug: 'index-astro', meta: { total_score: 20 }, body: 'open', cwd });
    const r = spawnSync(process.execPath, [SCRIPT, 'close', 'index-astro'], {
      cwd,
      encoding: 'utf-8',
    });
    assert.equal(r.status, 0, `stderr: ${r.stderr}`);
    assert.equal(readLatestSnapshot('index-astro', { cwd }), null);
  });

  it('close subcommand exits 2 when no snapshot exists', () => {
    const r = spawnSync(process.execPath, [SCRIPT, 'close', 'never-written'], {
      cwd,
      encoding: 'utf-8',
    });
    assert.equal(r.status, 2);
  });
```

And a unit test beside `writeSnapshot + readLatestSnapshot` that `closeSnapshot` returns the path and leaves `readLatestSnapshot` null.

## Alternatives considered

1. **Frontmatter `resolved: true` (or per-finding status).** Keeps trend history on disk. `latest` and routing then have to filter. More surface for the model to forget to set.
2. **`impeccable doctor` flags a snapshot whose target has commits after `timestamp`.** Useful as a mention, but doctor does not read critiques today, and a mention still leaves polish inheriting the file until someone acts.

Delete-on-close is the one `latest` already understands. Trend history is the last five snapshots that have not been closed; a closed run is one you can re-run.

Happy to open a PR if you would like it — following the issue-first rule and waiting for your go-ahead.

abdulwahabone:
Critique snapshots had no close path, so a leftover file under `.impeccable/critique/` was inherited by polish and the routing menu as a live backlog.

Adds a `close` subcommand to `critique-storage.mjs` that deletes every snapshot for the slug, and updates critique/polish/routing: critique leaves the file it just wrote, polish closes it when the target has moved on or once the inherited Priority Issues are cleared, routing points at polish without a freshness test of its own.

Fixes #660

Validation: `node --test tests/critique-storage.test.mjs` (27 pass), `bun run test` (fail 0), `bun run build` (skill prose clean). Generated provider output intentionally omitted (source-first PR).

AI disclosure: prepared with AI assistance (Cursor agent), directed and reviewed by maintainer @abdulwahabone.

### DIFF
--- a/skill/reference/critique.md
+++ b/skill/reference/critique.md
@@ -1,6 +1,6 @@
 ### Purpose
 
-Resolve one stable target, run two independent assessments, synthesize a design critique, persist a snapshot, and ask the user what to improve next. The chat response is the primary deliverable; the snapshot is an archive/backlog for future commands.
+Resolve one stable target, run two independent assessments, synthesize a design critique, persist a snapshot, and ask the user what to improve next. The chat response is the primary deliverable; the snapshot is an archive of that run.
 
 ### Hard Invariants
 
@@ -103,7 +103,7 @@ Codex failure accounting: final Run Notes must include target slug, ignore list,
 
 Synthesize both assessments into a single report. Do NOT simply concatenate. Weave the findings together, noting where the LLM review and detector agree, where the detector caught issues the LLM missed, and where detector findings are false positives.
 
-The chat response is the primary user-facing deliverable. Present the full structured critique below in chat; do not replace it with a summary and a link. The persisted snapshot is only an archive/backlog for later commands.
+The chat response is the primary user-facing deliverable. Present the full structured critique below in chat; do not replace it with a summary and a link. The persisted snapshot is an archive of that run.
 
 <codex>
 Codex final-answer note: `$impeccable critique` produces a report artifact, so the final chat response should intentionally exceed the usual concise close-out style. Do not title the final response "Critique Summary" unless the user explicitly asked for a summary.
@@ -231,7 +231,7 @@ Skip this step if the Setup slug was null (vague or root-level target).
    IMPECCABLE_CRITIQUE_META='{"target":"<user phrasing>","total_score":<n>,"max_score":<n>,"na_heuristics":"<comma-separated numbers, or empty>","p0_count":<n>,"p1_count":<n>}' \
      node {{scripts_path}}/critique-storage.mjs write "<resolved target>" <body-file>
    ```
-   `max_score` is the applicable maximum from the heuristic table (40 when every heuristic applied), so a later run can tell a renormalized total from a full one. The helper prints the absolute path it wrote.
+   `max_score` is the applicable maximum from the heuristic table (40 when every heuristic applied), so a later run can tell a renormalized total from a full one. The helper prints the absolute path it wrote. Leave that file on disk. Polish closes it; this run does not.
 
 3. **Delete the temp body file** after the write attempt completes, whether the write succeeded or failed. If deletion fails, mention `temp-file cleanup failed: <reason>` briefly in the final output, but do not block the critique.
 
--- a/skill/reference/critique.md
+++ b/skill/reference/critique.md
@@ -231,7 +231,7 @@ Skip this step if the Setup slug was null (vague or root-level target).
    IMPECCABLE_CRITIQUE_META='{"target":"<user phrasing>","total_score":<n>,"max_score":<n>,"na_heuristics":"<comma-separated numbers, or empty>","p0_count":<n>,"p1_count":<n>}' \
      node {{scripts_path}}/critique-storage.mjs write "<resolved target>" <body-file>
    ```
-   `max_score` is the applicable maximum from the heuristic table (40 when every heuristic applied), so a later run can tell a renormalized total from a full one. The helper prints the absolute path it wrote. Leave that file on disk. Polish closes it; this run does not.
+   `max_score` is the applicable maximum from the heuristic table (40 when every heuristic applied), so a later run can tell a renormalized total from a full one. For a local file target, the helper also records an exact content fingerprint so polish can distinguish the assessed bytes from later edits without relying on Git state or timestamps. The helper prints the absolute path it wrote. Leave that file on disk. Polish closes it; this run does not.
 
 3. **Delete the temp body file** after the write attempt completes, whether the write succeeded or failed. If deletion fails, mention `temp-file cleanup failed: <reason>` briefly in the final output, but do not block the critique.
 
--- a/skill/reference/polish.md
+++ b/skill/reference/polish.md
@@ -32,7 +32,7 @@ If a prior critique exists, use it as one input:
 node {{scripts_path}}/critique-storage.mjs latest "<resolved target>"
 ```
 
-Exit 0 returns the latest snapshot. For a local file target, the helper compares the file's exact current content fingerprint with the fingerprint captured by critique. Unchanged staged, unstaged, or untracked content remains current; any byte change closes the backlog while preserving its trend history and exits 2. A URL target has no local fingerprint and remains current until explicitly closed. When current, incorporate relevant P0/P1 findings and name the snapshot read. Exit 2 means none exists or the target changed. Perform an independent pass either way.
+Exit 0 returns the latest snapshot. For a local file target, the helper compares the file's exact current content fingerprint with the fingerprint captured by critique. Unchanged staged, unstaged, or untracked content remains current; any byte change, deletion, or replacement with a non-file closes the backlog while preserving its trend history and exits 2. A URL target has no local fingerprint and remains current until explicitly closed. When current, incorporate relevant P0/P1 findings and name the snapshot read. Exit 2 means none exists or the target changed. Perform an independent pass either way.
 
 ## 3. Triage
 
--- a/skill/reference/polish.md
+++ b/skill/reference/polish.md
@@ -32,7 +32,7 @@ If a prior critique exists, use it as one input:
 node {{scripts_path}}/critique-storage.mjs latest "<resolved target>"
 ```
 
-Exit 0 returns the latest snapshot. If the target's last commit is after the snapshot `timestamp`, close it and do not inherit:
+Exit 0 returns the latest snapshot. Read the target's last commit as `TZ=UTC git log -1 --date=format-local:%Y-%m-%dT%H-%M-%SZ --format=%cd <target>`; it shares the snapshot `timestamp` format, so compare the two as plain strings. When the commit stamp is greater, close the snapshot and do not inherit:
 
 ```bash
 node {{scripts_path}}/critique-storage.mjs close "<resolved target>"
--- a/skill/reference/polish.md
+++ b/skill/reference/polish.md
@@ -32,7 +32,7 @@ If a prior critique exists, use it as one input:
 node {{scripts_path}}/critique-storage.mjs latest "<resolved target>"
 ```
 
-Exit 0 returns the latest snapshot. For a file target, read its last commit as `TZ=UTC git log -1 --date=format-local:%Y-%m-%dT%H-%M-%SZ --format=%cd <target>`; it shares the snapshot `timestamp` format, so compare the two as plain strings. When the commit stamp is greater, close the snapshot and do not inherit:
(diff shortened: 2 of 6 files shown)


pbakaus:
I addressed the remaining data-retention finding in f00ece05. Closing now marks the newest snapshot with closed state instead of deleting files; latest ignores closed backlogs, trend retains their score history, a later critique reopens the slug naturally, and cross-target lookup still finds other live backlogs.

Validation: node --test tests/critique-storage.test.mjs (30/30), bun run build, and bun run test:core outside the sandbox (815/815).

AI-assisted maintainer repair: implemented, reviewed, and validated by Codex under maintainer direction.

pbakaus:
The working-tree freshness gap is addressed in f84dcccf. File targets now run a literal target-scoped git status check before commit-time comparison; staged, unstaged, and untracked changes close the inherited backlog instead of applying stale Priority Issues. A regression test enforces both the command and ordering.

Validation: 32 focused tests passed and bun run build completed with all prose/build validators green.

AI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
The timestamp/status heuristic is replaced in e33e2df4 with an exact SHA-256 fingerprint of the local target bytes. Critique records the fingerprint when writing; polish inherits an unchanged dirty or untracked file, and any byte change—including a same-second edit—closes only the live backlog while retaining trend history. Legacy local snapshots without fingerprints fail safely as stale; URL behavior is unchanged.\n\nValidation: 37 focused storage/reference tests, `bun run build`, and the authoritative `bun run test:core` suite (821/821) all passed.\n\nAI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
The missing-target edge case is addressed in e61a6aa1. Concrete local targets now require an exact match with the stored fingerprint, so deletion, rename, unreadability, or replacement by a directory closes the live backlog while retaining its trend. Explicit slug-only reads and URL targets keep their intentional no-local-file behavior.\n\nValidation: 38 focused storage/reference tests and `bun run build` passed; the immediately preceding fingerprint commit also passed the full core suite (821/821).\n\nAI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
The extensionless-file ambiguity is removed in 8a2af372. Fingerprinted snapshots now persist the resolved local target identity; `latest` uses that identity plus current filesystem presence before treating a slug-shaped argument as a slug. Existing, changed, and deleted extensionless targets therefore follow the same exact-content freshness path as files with extensions.\n\nValidation: 39 focused storage/reference tests and `bun run build` passed.\n\nAI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
Same-second writes are now collision-safe in 5fc8e094. Snapshot creation uses atomic `wx` writes and a fixed-width `~0001` suffix on collision, so concurrent or same-second critiques cannot overwrite history and lexical ordering still makes the newest snapshot current. Existing filename and timestamp formats remain valid.\n\nValidation: 40 focused storage/reference tests and `bun run build` passed.\n\nAI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
The slug/path identity collision is fixed in 44e735eb. Freshness invalidation now requires a stored target path to match the resolved local path; legacy snapshots retain only the prior unambiguous path-marker fallback. A regression reproduces the exact `foo/bar` versus extensionless `foo-bar` collision and verifies the unrelated backlog remains live while changes to the recorded original still close it.\n\nValidation: 39 focused critique-storage tests, `bun run build`, and the full core suite (825/825) passed.\n\nAI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
The close race is fixed in 4eda2ebc. `latest --json` now returns the exact `snapshot_file` identity, and `close` requires that identity instead of reselecting the newest file by slug. A regression reads snapshot A, writes B, closes A, and verifies B remains live through both per-slug and cross-target reads. Stale-target invalidation also closes the exact snapshot it inspected.\n\nValidation: 45 focused storage/reference tests, `bun run build`, and the full core suite (829/829) passed. The provider-backed `test:skill-behavior` run was attempted but blocked by the execution environment because it would transmit repository test content to third-party model APIs; no bypass was used.\n\nAI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
The ambiguous legacy extensionless case is fixed in 44cb795a. A slug-shaped legacy lookup with no recorded target identity now exits 2 without mutating history and asks for an explicit `./path` or full URL; retrying with `./main` safely treats it as a file and closes the stale legacy backlog. New extensionless snapshots remain directly readable because they record `target_path`.\n\nValidation: 46 focused storage/reference tests, `bun run build`, and the full core suite (830/830) passed.\n\nAI-assisted maintainer repair: implemented and validated by Codex under maintainer direction.

pbakaus:
Pushed the final target-identity hardening in 55541947. Snapshots now carry a canonical file or URL identity; exact targets must match it, intentional bare-slug lookup remains supported, and a same-named local file collision is rejected without inheriting or closing another target’s backlog.

Validation: `node --test tests/critique-storage.test.mjs tests/skill-reference.test.mjs` (46/46), `bun run build`, and `bun run test:core` (830/830). The provider-backed skill-behavior suite remains intentionally unrun because this environment did not authorize sending repository/test content to third-party model APIs.

AI disclosure: this repair and validation were performed by Codex under maintainer direction.

pbakaus:
Addressed the remaining same-slug/multiple-target lookup failure in 1ea435bb. `latest` now selects the newest snapshot within the requested canonical target identity before applying freshness checks, so an older `foo/bar` backlog remains reachable even when a newer `foo-bar` snapshot shares its filename slug. The regression also verifies that closing the first exact snapshot leaves the second target open.

Validation: focused storage/reference suite 47/47, `bun run build`, and `bun run test:core` 831/831.

AI disclosure: this repair and validation were performed by Codex under maintainer direction.

pbakaus:
Addressed cross-target backlog discovery in 3aa2d78b. The aggregation now keeps the latest open/closed state per canonical target identity rather than per lossy slug, with legacy snapshots retaining one stream per slug. Closing a newer colliding target therefore no longer hides an older target’s active backlog.

Validation: focused storage/reference suite 48/48, `bun run build`, and `bun run test:core` 832/832.

AI disclosure: this repair and validation were performed by Codex under maintainer direction.

pbakaus:
Addressed URL identity collisions in fc348cd3. URL snapshots now use normalized origin plus path, preserving scheme and explicit non-default port while retaining trailing-slash normalization. A regression writes colliding-slug snapshots for HTTP, HTTPS, and HTTPS on port 8443, verifies exact lookup for all three, closes one, and confirms the other two remain open.

Validation: focused storage/reference suite 49/49, `bun run build`, and `bun run test:core` 833/833.

AI disclosure: this repair and validation were performed by Codex under maintainer direction.

pbakaus:
Addressed cross-target close ownership in 0058b246. `close` now reads the exact snapshot identity and requires it to match the supplied resolved target before mutating modern snapshots; legacy identity-less snapshots retain their historical behavior. The URL collision regression now verifies that attempting to close an HTTP snapshot with the HTTPS target exits 2 and leaves the HTTP backlog open.

Validation: focused storage/reference suite 49/49, `bun run build`, and `bun run test:core` 833/833.

AI disclosure: this repair and validation were performed by Codex under maintainer direction.

pbakaus:
Addressed the remaining legacy-stream thread in 5d0f29e9. Cross-target routing now suppresses identity-less snapshots once that slug has any identity-aware record, preventing a closed modern snapshot from resurrecting an unreachable legacy backlog. Legacy-only slugs retain their historical routing behavior.

Validation: focused storage/reference suite 50/50, `bun run build`, and `bun run test:core` 834/834.

AI disclosure: this repair and validation were performed by Codex under maintainer direction.

pbakaus (approved):
Maintainer review complete. The repair now preserves critique trend history while closing only the exact processed snapshot, tracks local content and normalized URL identity, keeps colliding target streams independent, and prevents legacy snapshots from resurfacing after a modern close.

Validation: 50 focused storage/reference tests passed; source-first build passed for all 17 providers; 834 core tests passed; GitHub CI and Cursor Bugbot passed; Greptile reported 5/5 with no blocking failure; zero unresolved review threads.

AI disclosure: this maintainer review and repair were performed with Codex AI assistance under @pbakaus direction.

PR #693, opened 2026-09-01
Fix: vendor static-HTML parsers so skill installs detect fully (#434)

### LINKED ISSUE #434: [Bug] Skill install has no package.json, so the static-html engine never loads and detect silently degrades
### Summary

Installing impeccable **as a skill** — copying `.claude/skills/impeccable/` into a skills directory, which is how skills are distributed — leaves the static-html engine unable to load, so every `.html` scan silently degrades to `detectText`. The full CSS cascade never runs, and nothing tells the user.

Related to #359, but separable: that issue tests with the parsers **installed** and asks why a rule doesn't fire on the parsed DOM. This one is that in the normal skill-install path the parsers aren't there at all, so the engine never runs in the first place. The two share a remedy — #359's point 2 (surface the active engine) and the warning suggested below are the same idea.

### Cause

`scripts/detector/engines/static-html/detect-html.mjs` imports four packages at runtime:

```js
const [htmlparser2, cssSelect, csstree, domutils] = await Promise.all([
  import('htmlparser2'), import('css-select'), import('css-tree'), import('domutils'),
]);
```

The root `package.json` declares all four correctly, so `npm install` at the repo root is fine. But **the skill directory has no `package.json` of its own**, and copying `.claude/skills/impeccable/` brings the scripts without the dependencies. `import()` throws `ERR_MODULE_NOT_FOUND`, the bare `catch` at `detect-html.mjs:144` swallows it, and the scan quietly becomes a weaker scan.

Verified on upstream `main` (v4.0.2): the bare catch is still there, and no `package.json` exists under any of the `.<agent>/skills/impeccable/` trees.

### Why it is easy to miss

The fallback's gradient-text rule tests for the literal substring `gradient` within ±200 characters of `background-clip: text`. Detection therefore depends on what a CSS variable happens to be **named**:

```css
/* caught — the variable name contains "gradient" */
:root { --primary-gradient: linear-gradient(90deg,#7C3AED,#EC4899); }
.hero h1 { background: var(--primary-gradient); background-clip: text; color: transparent; }

/* missed — identical CSS, different variable name */
:root { --grad: linear-gradient(90deg,#7C3AED,#EC4899); }
.hero h1 { background: var(--grad); background-clip: text; color: transparent; }
```

The full engine resolves custom properties and catches both. The consequence is perverse: the better-organised a codebase's CSS, the less the fallback sees — a project using design tokens evades the rule it most needs.

### Reproduction

```bash
# skill-install shape: scripts present, no node_modules alongside
cp -r .claude/skills/impeccable /tmp/skill && cd "$(mktemp -d)"
printf ':root{--grad:linear-gradient(90deg,#7C3AED,#EC4899)}\n.hero h1{background:var(--grad);-webkit-background-clip:text;background-clip:text;color:transparent}' > s.css
printf '<html><head><link rel=stylesheet href=s.css></head><body><div class=hero><h1>Hi</h1></div></body></html>' > p.html

node /tmp/skill/scripts/detect.mjs p.html    # no finding, no indication a weaker analyser ran
```

Installing `htmlparser2 css-select css-tree domutils` next to the skill makes the same file report `gradient-text`.

### Measured impact

Same files, fallback vs full engine:

| File | Fallback | Full engine |
|---|---|---|
| Fixture built from documented anti-patterns | 2 | **8** |
| Generated page A | 3 | **6** |
| Generated page B | **0** | **5** |

Page B reported a clean zero while carrying five findings, two of them gradient text. Also recovered once the parsers resolve: `low-contrast` with computed ratios, `skipped-heading`, `dark-glow`, `ai-color-palette` — none of which the fallback can produce, since they need computed styles.

### Suggested fix

The one that matters is making the fallback announce itself. A checker that quietly becomes a worse checker is the most expensive failure it can have: a clean report is indistinguishable from an unanalysed one, so users act on a number that doesn't mean what they think it means.

```js
} catch (err) {
  if (!globalThis.__impeccableStaticHtmlWarned) {
    globalThis.__impeccableStaticHtmlWarned = true;
    process.stderr.write(
      'impeccable detect: DEGRADED — HTML parser modules unavailable ' +
      '(htmlparser2, css-select, css-tree, domutils).\n' +
      '  Falling back to regex matching. Custom properties, selector matching and computed ' +
      'contrast are NOT evaluated; findings are an undercount, not a clean bill of health.\n'
    );
  }
  return detectText(html, filePath, options);
}
```

Beyond that, two options that remove the failure rather than report it: ship a minimal `package.json` inside each skill directory declaring the four runtime dependencies, or vendor the parsers so the skill-install path is genuinely self-contained.

Happy to open a PR for the warning if useful.

abdulwahabone:
Skill and plugin copies of the detector imported htmlparser2 and friends from npm, so HTML scans silently fell back to regex and exited 0. This ships a self-contained parser bundle with the engine and exits 1 if that fallback still fires. Fixes #434.

AI assistance: Cursor Grok 4.6 implemented this change. Detector suite 182/182, core suite pass, bun run build copies the vendor file into dist skill trees.


abdulwahabone:
Not so sure about this fix, closing it temporarily

PR #713, opened 2026-09-03
Fix: ship static HTML parsers with skill installs

### LINKED ISSUE #434: [Bug] Skill install has no package.json, so the static-html engine never loads and detect silently degrades

pbakaus:
## Summary

- bundle the static-HTML parser stack inside every copied skill and plugin install, so scans no longer depend on a nearby `node_modules`
- generate the bundle with exact-pinned esbuild and fail builds when either the parser artifact or its licenses are stale
- ship complete license text for all twelve bundled packages and keep the multiplied provider artifact below a 256 KB ceiling
- preserve #711’s operational-failure semantics: a missing bundle still returns fallback findings and valid JSON, but exits 1

This supersedes #693 while preserving @abdulwahabone’s original commits and authorship. Generated root harness output is intentionally omitted; the post-merge sync workflow will refresh it.

Closes #434.

## Validation

- `node --test tests/detect-static-html-skill-install.test.mjs`
- `bun test tests/lib/detector-bundle.test.js`
- `bun run test:detector` (442 Bun tests + 186 Node/Chromium tests)
- `bun run test` (complete default suite, exit 0)
- `bun run build`
- `node scripts/build-static-html-parsers.js --check`
- `npm pack --dry-run --ignore-scripts --json` (both generated parser files included; 430 KB packed)
- verified parser bundle and license file in all 37 generated detector copies

AI assistance disclosure: Codex audited #693, implemented the hardening changes, and ran the validation above.

---

> [!NOTE]
> **Medium Risk**
> Changes how static-HTML detection runs in production installs; a bad or stale bundle could silently degrade rule coverage until CI `--check` catches it.
> 
> **Overview**
> **Ships the static-HTML parser stack inside the detector** so skill and plugin installs work without a local `node_modules`. Runtime loads a committed **`static-html-parsers.mjs`** bundle (htmlparser2, css-select, css-tree, domutils) instead of dynamic npm imports.
> 
> Adds an **esbuild-based generator** with `--check` freshness gates wired into **`build`** / **`build:release`**, a **256 KB size cap**, and **`static-html-parsers.LICENSES.txt`** for all bundled packages. **NOTICE.md** documents the vendored stack.
> 
> When the bundle is missing, behavior stays **degraded regex fallback** with stderr warning, but the CLI now records **`onOperationalFailure`** (parser-bundle-unavailable) so scans can **exit 1** while still emitting valid JSON—aligned with existing operational-failure semantics.
> 
> New tests cover skill/plugin-shaped install trees, stale-bundle CI checks, and vendor inclusion in the skill detector copy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff94918ddc2b29a5f6f4d0e2812034827b0516dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pbakaus:
Closing as superseded by the Rust engine merged in #714, which embeds static HTML parsing and removes the JavaScript dependency problem this PR addressed. Thanks to @abdulwahabone for the original approach in #693. #434 remains open until fresh provider and marketplace-plugin installations are verified.

AI-assisted maintainer update prepared with Codex.

Issue #744, opened 2026-09-06
[Bug] Follow-up to #736: 4.2.1 activates, but nothing behind the activation runs — 0 of 35 reference files read, 55 of 55 launcher calls refused


ktlesr:
## What happened?

First, thank you — #738 fixes #736 completely. In 4.2.1 the skill activated in
**28 of 28** attempts where the model reached for it, with zero refusals, so I
was able to run the full trigger-discrimination measurement that #736 blocked.

The headline is good news and I want to lead with it: **precision 100% (N=28,
95% CI 88%–100%), zero false positives in 70 negative attempts.** Seven negative
cases — four of them frontend work with no visual decision in it, one of those
in the same file a positive case targets; two of them a real visual decision
that is not an interface — and the boundary your description declares held on
every single one.

This issue is about what the measurement found *downstream* of activation. Two
findings are much stronger than the rest, so they go first.

### 1. The completion case: 0/10 activations, 10/10 `DESIGN.md` written

One case asks, in the skill's own terms but without naming any sub-command or
using the word "design":

> Work out the visual system this codebase is already using — type scale,
> colour, spacing, component patterns — and write it down at the project root,
> in the file a later session would look for, so a new contributor can pick it
> up without asking.

That is what `document` promises. Across ten attempts:

| | |
| --- | --- |
| activations | **0 / 10** |
| `DESIGN.md` written at project root | **10 / 10** |

No `Skill` call appears in any of the ten. The base model produced the artefact
the skill exists to produce, at the path the skill documents, and
`reference/document.md` — 27,844 bytes specifying the DESIGN.md format, its
token schema and its fixed eight-section order — was never opened. So the ten
files that got written are the model's idea of a design system, not yours.

From a user's side this looks like a success. From an author's side the format
spec never ran.

### 2. The 35 reference files are not being read

| | Files | Bytes |
| --- | --- | --- |
| ships under `reference/` | 35 | 352,712 |
| opened across 28 activations | **0** | **0** |

**One** reference file was reached for in 28 activations — `onboard.md`, once —
and the host refused that read because the plugin directory sits outside the
session workspace. The other 27 activations never named a reference file at all.

The specific consequence I would look at first: `SKILL.md` step 3 says to load
`reference/craft-floor.md` **immediately before editing UI**, and calls it the
quality floor and the absolute bans. All 28 activations edited UI. None loaded
it.

I do not think these two findings are independent of finding 3 below, and the
[...]

Issue #375, opened 2026-07-16
Proposal: slim the skill description (~120 words loaded into every context window)


lukemavdynamics:
Hi Paul! First off, I love impeccable! I use it all the time for frontend development and it has made me far more confident in developing front end (so thank you!)

One thing I noticed while auditing the always-loaded skill descriptions in my setup: Impeccable's `description` in [skill/SKILL.src.md](https://github.com/pbakaus/impeccable/blob/main/skill/SKILL.src.md) is around 119 words, which makes it the largest single always-loaded item in my harness by roughly 3x. Since harnesses inject every skill description into every context window, that cost is paid in every session, including ones that never touch UI work.

My read is that the enumeration style (the "design, redesign, shape, critique, audit, polish..." verb list, plus the surface and topic lists) dates from when Impeccable shipped ~20 separate skills that each needed their own trigger terms to compete for selection (the https://github.com/pbakaus/impeccable/pull/55 era, it seems). Now that everything is consolidated into one skill, category words should cover the same trigger space, and the sub-command vocabulary already lives in `argument-hint` and the command table inside the skill body.

Here's a ~45-word sketch of what I mean:

> Design, build, or refine any frontend interface: layout, typography, color, motion, accessibility, responsive behavior, UX copy, design systems. Use when the user wants UI designed, critiqued, or improved, asks to make a design bolder or quieter, or invokes an impeccable sub-command. Not for backend-only tasks.

I might be missing selection-accuracy data that justifies the longer form, so totally open to this being wrong. Happy to send a PR if you think it's worth pursuing.

abdulwahabone:
@lukemavdynamics Thanks for filing this issue.
We should hold this off until we have some proper evals to test the change in the output.

pbakaus:
Linking #744, which now provides activation measurements and a pending 4.2.2 rerun. Let’s use that case set to compare a shorter description against the current one, including document-shaped requests and non-UI negatives. The goal is fewer tokens without losing useful activation or introducing false positives. Holding implementation until that comparison.

AI assistance: Codex, under maintainer direction.

Issue #789, opened 2026-09-09
[Feature] Degraded Setup path that auto-loads craft-floor + reference pack when launcher is refused


mvanhorn:
## What problem does this solve?

Issue #744 measured Claude Code 4.2.1 activation: precision 100%, but **0 of 35 reference files** were read across 28 activations, `craft-floor.md` never loaded before UI edits, and Setup step 1 (`impeccable context`) was refused 55/55 times in non-interactive hosts. The model then improvises from `SKILL.md` alone — including writing `DESIGN.md` without ever opening `document.md` (0/10 Skill calls, 10/10 files written).

Users want the *authored* playbooks (craft floor + command-specific refs), not a generic model approximation. When the launcher cannot run, there is currently no permission-safe path that still injects Impeccable's craft guidance before UI edits.

## Proposed solution

Add an explicit **degraded Setup path** in `skill/SKILL.src.md` + `reference/init.md`:

1. After Setup step 1 (`impeccable context` / launcher) fails or is permission-denied, do **not** silently continue.
2. Emit a short user-visible degraded notice (what failed, what will still load).
3. **Must-read pack** before any UI edit: `reference/craft-floor.md` (always), plus command-specific refs (`document.md` when writing DESIGN.md, `craft.md`/`new-work.md` for build flows).
4. Document the degraded contract in `docs/CLI-CONTRACT.md` and skill-behavior scenarios so hosts without launcher consent still get authored guidance.

Out of scope for this proposal: changing host permission models; expanding #375 description slim (related but separate).

Source-first edits under `skill/` only (no generated harness churn in the feature PR). Add a skill-behavior scenario asserting: when launcher stub returns permission error, the agent Read-calls `craft-floor.md` before Edit on UI files. Do **not** bump package/plugin/extension versions or changelog.

### Acceptance criteria

- [ ] SKILL/init document a degraded Setup path with must-read reference pack when launcher fails.
- [ ] `craft-floor.md` is required before UI edits even when Setup step 1 is refused.
- [ ] `document.md` is required before writing DESIGN.md in the completion/document case.
- [ ] skill-behavior (or equivalent) scenario covers launcher-denied degraded path.
- [ ] No version bump / changelog / unsolicited generated harness sync in the feature PR.

## Provider(s) this applies to

- [ ] Cursor
- [x] Claude Code
- [ ] Gemini CLI
- [ ] Codex CLI
- [ ] VS Code Copilot
- [ ] Grok Build
- [ ] Kiro
[...]

PR #791, opened 2026-09-09
test: scenario 19 documentation case when the context launcher is denied

mvanhorn:
## Summary

Most of the degraded Setup path already shipped on main (#750, #756, #788). This PR keeps the remaining coverage: a scenario 19 documentation case that requires `document.md` and the incumbent source to be read before DESIGN.md is written when the context launcher is denied, plus the README measurement for that case.

The documentation case also passes on main's unchanged `SKILL.src.md` and `init.md` with this test file, so the coverage is the test.

#744's completion case (skill never loaded, DESIGN.md still written) is an activation failure. That belongs with #375. A must-read pack cannot reach a run that never loaded the skill, and refused reference reads outside the workspace fail the same way.

## Changes

- `tests/skill-behavior/scenarios.test.mjs` — denied-launcher documentation case. The degraded notice must land before the first tool call after the denied launcher.
- `tests/skill-behavior/README.md` — scenario 19 row and 2026-09-09 measurements: 4/4 on claude-sonnet-5, 4/4 on gpt-5.6-terra, 4/4 on gemini-3.7-flash (one run each).

No version bump, no changelog, no generated harness sync. No skill-text change.

## Demo

HyperFrames walkthrough (45s): launcher refused → degraded notice → craft-floor must-read → DESIGN.md only after document.md.

Degraded Setup path walkthrough

<video src="https://raw.githubusercontent.com/mvanhorn/impeccable/walkthrough-assets/degraded-setup-walkthrough.mp4" controls muted width="960"></video>

[Full-res MP4](https://raw.githubusercontent.com/mvanhorn/impeccable/walkthrough-assets/degraded-setup-walkthrough.mp4)

## Notes

AI was used for assistance.


pbakaus (changes requested):
Thanks for this, Matt. I ran the branch through the build and the skill-behavior suite, and the direction is sound. The main thing I'd change is scope: most of the degraded path already shipped, so the prose parts of this PR say things twice.

**What's already on main.** 4.2.2 (#750) added the "Launcher unavailable" fallback: notice before the next tool call, read PRODUCT.md and DESIGN.md directly, then follow steps 2 and 3. Step 2 is the command reference (so `document.md` for `/impeccable document`), step 3 is craft-floor. 4.2.3 (#756) tightened the wording and added scenario 19. 4.3.1 (#788) fixed the launcher's own silent failures. That covers four of the five acceptance criteria in #789 before this PR.

**Measurements.** Scenario 19 on this branch, all four cases: 4/4 on claude-sonnet-5, 4/4 on gpt-5.6-terra, 4/4 on gemini-3.7-flash. I then reran your new documentation case with main's unchanged `SKILL.src.md` and `init.md` and your test file: it passes on all three providers too. So the test is real coverage, but the text change is not what earns it.

**On the #744 framing.** The completion case there (0/10 activations, 10/10 DESIGN.md written) is an activation failure: SKILL.md was never loaded in those runs, so nothing inside it can reach the model. That's #375's problem. The refused reference reads in #744 (plugin dir outside the workspace) would defeat a must-read pack the same way. Worth keeping those separate in the PR description.

**Requested changes:**

- Remove `AI_PR_NOTICE.txt`. I'm fine with the PR, but the notice file shouldn't land in the tree.
- Drop the `skill/SKILL.src.md` change. The paragraph now restates steps 2 and 3, and every session pays for those tokens. If you want an anchor for the test, the rule marker alone is fine.
- Drop the `init.md` paragraph. It restates the file's opening sentence and the "never offer DESIGN.md during init" line.
- Drop the `docs/CLI-CONTRACT.md` section. That doc is the verb contract, byte for byte, and this describes behavior with no CLI output. It also lists "missing binary" as a degraded trigger, but since 4.3.1 that's the launcher running and printing, which the section's own last sentence puts in the other bucket.
- Keep the scenario 19 test and the README row. Replace the "not measured on any provider yet" note with the numbers above (one run each, 2026-09-09).

Build and both prose validators pass on the branch as is.

Review drafted with AI assistance (Claude Code), verified and posted by me.

mvanhorn:
Addressed the requested changes on `b72fb2a5`.

- Removed `AI_PR_NOTICE.txt`.
- Reverted `skill/SKILL.src.md`, `skill/reference/init.md`, and `docs/CLI-CONTRACT.md` to match main. No restated prose. The documentation test does not need a rule-marker anchor, so that file is unchanged from main.
- Kept the scenario 19 documentation case and the README row. Measurements are now the 2026-09-09 numbers: 4/4 on claude-sonnet-5, 4/4 on gpt-5.6-terra, 4/4 on gemini-3.7-flash, one run each. The README also notes the documentation case passes on main's unchanged skill text.
- Tightened the documentation assertion so the degraded notice is compared with the first tool call after the denied launcher, not only before the eventual write.

Greptile P1 is moot with the `SKILL.src.md` revert. PR description no longer treats #744 activation failures as something this change can fix.

AI was used for assistance.

pbakaus (approved):
thanks!

PR #111, opened 2026-04-26
Rename all references to `Neon Mirai` -> `Neo Mirai`, to match case study site's actual content

vivshaw:
## Summary

In 74f16d6310ea0e49d1032d4ed51b0615bb78524f, [the Neo Mirai case study](https://impeccable.style/cases/neon-mirai/) was added. However, most references in the docs called it `Neon Mirai`. This appears to have been a typo, as all visual mockups and the live case study clearly say `Neo Mirai`, not `Neon`.

This PR simply batch-renames to `Neo Mirai`, to match the actual content of the case study. Since this required editing live URLs that have been up for a few days, I included redirects for the affected routes.

I've tested this by firing up the server locally and poking through the routes.

## Type of change

- [ ] New command
- [x] New / updated skill reference
- [ ] New anti-pattern guidance
- [x] Bug fix
- [x] Documentation update
- [ ] Build system / tooling
- [ ] Other: 

## Checklist

- [x] Source files updated in `source/`
- [x] `bun run build` ran successfully
- [ ] `bun test` passes
- [ ] Tested with at least one provider (Cursor / Claude Code / Gemini CLI / Codex / Copilot / Kiro / OpenCode)
- [ ] README / DEVELOP.md updated if needed

---

> [!NOTE]
> **Low Risk**
> Low risk copy/URL consistency update; the only functional change is adding redirects for legacy `neon-mirai` paths, which could affect navigation if mis-typed but is straightforward.
> 
> **Overview**
> Renames the case study reference from **“Neon Mirai” to “Neo Mirai”** across skill reference docs and site pages, updating titles, descriptions, alt text, and internal links to use the `neo-mirai` routes.
> 
> Adds permanent redirects for legacy `/neon-mirai` and `/cases/neon-mirai` URLs in both the build-generated Cloudflare `_redirects` file and the dev server router, and updates the build pipeline to include the `neo-mirai` case pages plus the sitemap canonical entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 579006cda5cfa55465835a87e84ff0c223bd9a80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #798, opened 2026-09-09
Fix detect DOM routing for template and SFC extensions

### LINKED ISSUE #795: [Bug] `detect` routes only .html/.htm to the DOM engine, so contrast/a11y rules never run on Blade, Vue, Svelte or configured template extensions
## What happened?

`impeccable detect` routes a file to the DOM-aware engine only when its extension is exactly `.html`/`.htm`. Every other markup-bearing extension — including built-in ones like `.blade.php`, `.vue`, `.svelte`, and any `detector.extensions` entry such as `.html.erb` — is sent to the text engine instead.

The files *are* scanned; they are misrouted. Text-engine rules still fire, so the file looks covered. But DOM-dependent rules (contrast, tap targets, heading hierarchy, ARIA) never run, and `detect` exits 0 with `[]`. A skipped dimension is indistinguishable from a passing one.

The hook path does **not** have this bug — it routes the same file correctly. So the same file scanned two ways gives two different answers.

This looks like the same class of path-divergence as #460, but on the `detect`-vs-hook axis rather than file-vs-directory.

## Steps to reproduce

Identical bytes, two extensions, no config needed:

```sh
BODY='<a href="#" style="color:#ccc;background:#fff">low contrast</a>'
printf '%s\n' "$BODY" > page.html
printf '%s\n' "$BODY" > page.blade.php   # built-in extension
printf '%s\n' "$BODY" > page.html.erb    # via detector.extensions

impeccable detect --json page.html        # -> low-contrast finding
impeccable detect --json page.blade.php   # -> []
impeccable detect --json page.html.erb    # -> []
```

Proof the file is scanned but misrouted — a text-engine rule fires on `.vue` while a DOM rule does not:

```sh
printf '<template>\n  <div style="transition: all 0.3s cubic-bezier(0.68,-0.55,0.27,1.55)">y</div>\n</template>\n' > t.vue
cp t.vue t.html
impeccable detect --json t.html   # -> ['bounce-easing']
impeccable detect --json t.vue    # -> ['bounce-easing']   (text rule: fires)
# but the contrast case above returns [] for .vue and a finding for .html
```

And the hook disagrees with `detect` on the very same file:

```sh
cat > .impeccable/config.json <<'JSON'
{"detector":{"extensions":[{"ext":".html.erb","engine":"html"}]}}
JSON

impeccable detect --json page.html.erb   # -> []
printf '{"tool_name":"Edit","tool_input":{"file_path":"'$PWD'/page.html.erb"},"cwd":"'$PWD'"}' \
  | impeccable hook                      # -> reports the low-contrast finding
```

`impeccable doctor` reports "No drift found" throughout.

## Expected

`detect` routes to the DOM engine for any extension that emits markup — the built-in list plus configured `engine: "html"` entries — and agrees with the hook on the same file.

## Likely cause

`crates/detect/src/file_system.rs`:

```rust
pub const HTML_EXTENSIONS: &[&str] = &[".html", ".htm"];

pub fn is_html_path(file_path: &str) -> bool {
    HTML_EXTENSIONS.contains(&to_lower_case(&jsp::extname(file_path)).as_str())
}
```

`extname` returns the last segment only, so `show.html.erb` yields `.erb` and `page.blade.php` yields `.php`. Neither is in `HTML_EXTENSIONS`. Note `has_scannable_extension` directly above it already handles multi-part extensions with an `ends_with` loop — `is_html_path` appears not to have received the same treatment, and does not consult configured extensions at all.

## Impact

Any CI step or script calling `impeccable detect` on a Rails, Laravel, Vue, or Svelte project reports a clean pass over files whose accessibility rules were never evaluated. The failure is silent and green.

## Environment

- skill 4.3.1, engine 0.1.5
- macOS arm64 (Darwin 25.5.0)

mvanhorn:
Fixes #795.

## Problem

`impeccable detect` sent a file to the DOM engine only when last-segment `extname` was `.html` or `.htm`. Built-in markup (`.vue`, `.svelte`, `.astro`, `.blade.php`) and configured `detector.extensions` with `engine: "html"` (for example `.html.erb`) were scanned by the text engine instead.

The files looked covered: text-engine rules still fired. Contrast, ARIA, tap-target, and other DOM rules never ran, so `detect --json` exited 0 with `[]`. The hook already suffix-matched configured extensions, so the same file produced two answers.

## Solution

Shared routing helper in `crates/detect/src/engine_route.rs`, used by both `detect` and the hook:

- Multi-part suffixes via `ends_with` on the basename (same pattern as `has_scannable_extension` / `match_configured_extension`).
- Built-in DOM set: `.html`, `.htm`, `.vue`, `.svelte`, `.astro`, `.blade.php`.
- `detector.extensions` from `.impeccable/config.json` and `config.local.json`: `engine: "html"` → DOM, `engine: "text"` → text. Longest suffix wins; a configured entry of equal length overrides the built-in.
- Directory walks include configured suffixes so `detect .` covers `.html.erb` once declared.
- Hook allowlist now treats built-in DOM suffixes as in-scope without a config entry. Configured overrides stay authoritative.

Docs: `docs/CLI-CONTRACT.md` and `skill/reference/hooks.md` now describe the same extension→engine map.

## Test plan

- Fixture pack `tests/fixtures/detect/template-dom-routing/` with identical low-contrast anchor bytes as `page.html`, `page.vue`, `page.svelte`, `page.blade.php`, `page.html.erb`.
- `crates/detect/tests/template_dom_routing.rs`: `detect --json` yields the same low-contrast family on every DOM-routed path; unconfigured `.html.erb` does not; a directory walk includes configured `.html.erb`.
- Hook parity in `crates/hook/tests/hook_tests.rs`: `.vue` / `.blade.php` hit the HTML engine without config; `.html.erb` needs `engine: "html"`.
- Ran `cargo test -p impeccable-detect` and `cargo test -p impeccable-hook` (also with `RUSTFLAGS="-D warnings"`).

## Walkthrough

HyperFrames walkthrough (42s): on main, `detect --json` finds low-contrast in `.html` but returns `[]` for the same markup in `.vue`. This PR's binary matches `.html`, `.vue`, and configured `.html.erb`.

Detect DOM-routes template extensions walkthrough

<video src="https://raw.githubusercontent.com/mvanhorn/impeccable/walkthrough-assets/detect-dom-route-template-extensions.mp4" controls muted width="960"></video>

[Full-res MP4](https://raw.githubusercontent.com/mvanhorn/impeccable/walkthrough-assets/detect-dom-route-template-extensions.mp4)

## AI disclosure

Prepared with AI assistance (Cursor Grok).

---

> [!NOTE]
> **Medium Risk**
> Changes core detect/hook routing, hybrid finding merge, and hook cache versioning, but behavior is heavily regression-tested and mostly aligns previously divergent paths.
> 
> **Overview**
> **Detect and the design hook now share one extension→engine map** (`engine_route`), so `.vue`, `.svelte`, `.astro`, `.blade.php`, and configured `detector.extensions` with `engine: "html"` (e.g. `.html.erb`) get the same DOM analysis as plain HTML instead of text-only routing on last-segment `extname`.
> 
> **Hybrid template scanning** runs the full text pipeline plus DOM checks for non-`.html` markup; text findings keep line numbers and cache keys, and DOM-only rules are added when text does not already report them. The static HTML engine gains **`detect_html_source`** (used by Cursor pre-write without temp files), conservative computed-style checks when classes or preprocessor blocks are unresolved, and source-line attribution on DOM findings.
> 
> **Config and walks:** `detector.extensions` is merged in detect config; directory scans honor configured HTML suffixes and skip **`vendor`**. Hook scan expansion, Stop baselines, and mixed **`classify_mixed`** attribution align with markup templates; hook cache **version 2** drops stale `rule:0:snippet` sessions. CLI help, `CLI-CONTRACT.md`, hooks docs, fixtures, and oracle goldens are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54f8655d160b560b564e1ab33fea33cc6c094de2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pbakaus (changes requested):
Thanks for taking #795 on. The diagnosis in the PR body is right, but the fix as written is an either/or engine swap, and the DOM engine is not a superset of the text engine for component files. As it stands the PR removes more coverage than it adds for `.vue`, `.svelte`, `.astro`, and `.blade.php`, and it breaks the oracle. Requesting changes.

Reviewed at efe3250. Findings were verified by building the PR head and comparing against a main build and the released 4.0.0 engine, on the repo's own fixtures and small temp projects.

## Regressions in what `detect` catches

**The entire text-rule family disappears for SFC and template files** (`crates/detect/src/engine_route.rs:17`, confirmed). The HTML engine runs only three text-content analyzers, gated on `is_full_page` (`crates/html/src/engine.rs:254`), and never runs `run_regex_matchers`, `extract_style_blocks`, `scan_inset_stripe_css`, pseudo-stripe, or codex-grid-background. Surviving findings also lose their line numbers. Measured PR-head vs 4.0.0 on the repo's fixtures:

| fixture | before | after |
|---|---|---|
| `astro-inset-shadow-stripe.astro` | 17 side-tab, exit 2 | 0, exit 0 |
| `pseudo-stripe.vue` | 1 | 0 |
| `vue-should-flag.vue` | 8 with line numbers | 4 at line 0 |
| Card.vue with `<style lang="scss">` nested rules | 5 | 0 (the CSS parser cannot cascade SCSS) |

`impeccable detect src/` on a Nuxt or SvelteKit app now prints nothing and exits 0 for exactly the patterns it exists to catch, and the per-edit hook (`hook.rs:235`, `before_edit.rs:798`) has the same hole. The fix at the right depth: for markup-bearing non-`.html` files run `detect_text` as well and merge with dedupe, or have the HTML adapter run the full text pipeline for SFC-routed sources instead of only the three analyzers.

**The DOM engine has no utility-class resolution, so the should-pass fixtures now flag** (`crates/detect/src/cli.rs:277`, confirmed). With the PR binary, `detect --json --no-design-system` on `vue-should-pass.vue` and `svelte-should-pass.svelte` each emit `cramped-padding: <div> "card": children flush against bg on all sides (no inset)` and exit 2 (main: `[]`, exit 0), because `p-6` / `p-4` are classes with no cascade behind them. `svelte-should-flag.svelte` gains a bogus `low-contrast 1.0:1 text #000000 on #000000` because `text-purple-500` is unresolvable and default black meets the `.sidebar` background. Both are immediate-tier rules, so every Tailwind-styled Vue/Svelte/Astro edit now interrupts the agent with a correction prompt, and a CI gate on detect's exit code starts failing on clean components. The only escape is an undocumented `{ext: ".vue", engine: "text"}` override.

**Stop-baseline attribution switches off for SFCs** (`crates/hook/src/hook.rs:235`, confirmed). `use_html_engine` is now true for these suffixes, and `stop_baseline::capture` returns early when it is (`stop_baseline.rs:81`), both reconcile calls sit under `if !use_html_engine` (`hook.rs:284`, `:714`), and classify passes None as the baseline (`stop_baseline.rs:215`). Reproduced with both binaries: temp project, `src/Card.vue` with a pre-existing `border-left: 4px solid #6366f1`, PostToolUse Edit then Stop on the same session. Main writes a `stopBaseline` entry and Stop emits nothing. PR head writes none and Stop emits `[side-tab] [attribution unknown] ...` plus the "may predate this session" note for a finding the session never introduced. A one-line copy fix in a component with legacy debt now dumps the file's backlog at Stop and prompts the agent to expand scope. The per-edit reminder also loses its `L<n>` marker for SFCs.

## Test gate

**Twenty oracle goldens fail and the PR neither re-records them nor adds a `DELTAS.md` entry** (`engine_route.rs:146`). CLAUDE.md: the oracle is the behavior gate, and verb-level changes in `crates/` run `IMPECCABLE_BIN="$PWD/target/release/impeccable" bun run test` and add an oracle case when the contract grows. Replayed with the PR build: 811 pass, 20 fail, all green on main:

```
detect-fixture-{json,text}-{astro-inset-shadow-stripe-astro,pseudo-stripe-vue,
  svelte-should-flag-svelte,svelte-should-pass-svelte,vue-should-flag-vue,vue-should-pass-vue}
detect-dir-text-all-fixtures (415 -> 392 anti-patterns)
detect-dir-quiet-all-fixtures
detect-scope-type, detect-scope-layout-text, detect-scope-both
detect-no-advisory-json, detect-no-advisory-text
detect-help
```

The test plan lists only `cargo test -p impeccable-detect` and `-p impeccable-hook`. `detect-dir-json-all-fixtures` is on the accepted-delta list, so part of the regression is masked rather than caught. The contract also grows (`detector.extensions` honored by `detect`, configured-suffix directory walks, `.blade.php` hook scope, new `--help` text) with no new oracle case.

## Scope and false blocks

**Configured extensions widen `detect <dir>` into dependency trees** (`crates/detect/src/file_system.rs:116`, confirmed). `walk_dir_reporting_with` now pushes every file matching a configured suffix regardless of engine, `SKIP_DIRS` has no `vendor` entry, and there are no default `ignoreFiles`. Reproduced: `.impeccable/config.json` = `{"detector":{"extensions":[{"ext":".php","engine":"text"}]}}` (the shape the PR's own `longest_suffix_wins_builtin_over_shorter_configured` test uses) plus `vendor/x/y.php` and `app/Foo.php`; `detect --json .` reports gradient-text for both, where before the walk was limited to `SCANNABLE_EXTENSIONS`. The string form `[".php"]` normalizes to engine html and DOM-parses every controller and migration as markup. A Laravel or Rails (`.html.erb` under `vendor/bundle`) project gets thousands of third-party files scanned and findings in code it does not own.

**The Cursor pre-write gate now DOM-scans Astro layouts from a temp dir** (`crates/hook/src/before_edit.rs:798`, confirmed). Proposed `.astro`/`.vue`/`.svelte`/`.blade.php` content goes through `detect_proposed_html`, which writes it to `$TMPDIR/impeccable-pre-<stamp>/<basename>`, so `<link rel="stylesheet" href="../styles/global.css">` resolves against the temp dir, fails silently into the sink (`hook_lib.rs:1628`), and the rules see a cascade missing the project's CSS. Reproduced: `src/styles/global.css` with `body{color:#111;background:#fff} .hero{color:#fff}` and `src/layouts/Base.astro` linking it with `<div class="hero" style="background:#000">`. PR binary: `detect --json src/layouts/Base.astro` = `[]`, but `hook-before-edit` with Cursor-shaped stdin returns `{"permission":"deny", ... [low-contrast] ...}`. Main returns `allow`. Pre-existing for `.html`, but Astro layouts are precisely the file class that carries `<link rel=stylesheet>`, so this becomes a routine false block on every layout edit in Cursor.

**The parity claim for `.blade.php` and `.html.erb` does not hold** (`crates/detect/src/detect_text.rs:34`). `run_text_content_analyzers` gates on `ext_from_file_path` (`\.\w+$`, yielding `.php` / `.erb`) being in `PAGE_ANALYZER_EXTS`. Reproduced: a full-page document with buzzword copy as `page.html` yields `marketing-buzzword (8 buzzword phrases)`; the same bytes as `page.blade.php`, or as `page.html.erb` with `{"ext":".html.erb","engine":"html"}` configured, yield `[]`. The gap predates the PR, but `docs/CLI-CONTRACT.md` line 186, `skill/reference/hooks.md`, and `builtin_markup_and_configured_erb_share_the_html_finding_family` now assert parity that does not hold; the test checks a one-line fragment that never reaches the gate. The page-analyzer gate should consult the same suffix map instead of the last-segment regex.

## Cache and duplication

**Hook cache keys change shape across the engine flip** (`crates/hook/src/hook_lib.rs:958`). Text-engine findings key as `rule:line[:value]`, HTML-engine findings as `rule:0:snippet`, and `read_cache` validates only `version == 1` with no engine stamp, so nothing invalidates old `.vue`/`.svelte`/`.astro` keys. A session spanning the upgrade has `src/Card.vue` under `side-tab:20:...`; the next edit produces `side-tab:0:border-left: ...`, `dedupe_against_cache` misses every old key, and the agent gets acknowledged findings re-emitted as fresh instead of the pending-ack path. Secondary, now extended to SFCs: two same-rule findings at line 0 with identical 80-char snippets collapse to one key, and `hooks ignore-value` has no line to point at.

**`is_hook_scan_path` is a sixth hand-maintained extension list** (`hook_lib.rs:49`). `ALLOWED_EXTS` is `SCANNABLE_EXTENSIONS` minus `.blade.php`, which the middle term adds back, so the three-way OR equals `has_scannable_extension(p) || match_configured_extension(p, exts).is_some()`, re-implementing the walker predicate the same PR writes at `file_system.rs:116`. The lists already disagree: `.blade.php` is DOM-routed and hook-scanned but absent from `UI_CODE_EXTS` (`hook_lib.rs:2098`, keyed on extname), so unlike `.vue` it gets no co-located stylesheet co-scan. This drift is the exact shape of #795. Cheaper and deeper: one `is_scannable(path, &config.extensions)` in `impeccable-detect` used by `walk_dir`, `run_hook`, `run_stop_hook`, and `before_edit`, with the built-in lists derived from a single file-kind table.

**`upsert_ext` duplicates `JsMap::set`** (`engine_route.rs:82`). The find-then-replace-else-push body is line for line `JsMap::set` in `crates/foundation/src/js_ext_a.rs:161`, re-exported via `impeccable_core::js_ext_a`, which this crate already depends on and which `css_scan.rs` and `browser/driver.rs` already import. `merge_extensions` can build a `JsMap<ExtensionEntry>` and return its entries in order.

## Suggested direction

Keep the routing map, but make markup-bearing non-`.html` files run both engines and merge with dedupe (or run the full text pipeline inside the HTML adapter for SFC-routed sources). Keep the Stop-baseline mechanism on for those files. Re-record the goldens, review the diffs by hand, add a `DELTAS.md` entry for the intended changes and an oracle case for the new configured-suffix behavior. Scope the directory walk widening to the html engine or gate it on an explicit opt-in.

---
Review prepared with AI assistance (Claude Code), verified by a human maintainer before posting.

mvanhorn:
Thanks for the depth here, particularly the measured before/after table. You were right on all five counts, so I reworked the PR rather than patching it. Pushed in 7d30768.

- Built-in scan and routing suffixes derive from one file-kind table, and detect exposes a single `is_scannable(path, extensions)` used by the directory walk, `run_hook`, `run_stop_hook`, and `before_edit`, replacing the five hand-maintained lists.
- Markup-bearing non-`.html` sources run the full text pipeline and merge with the DOM findings, so the text-rule family and its line numbers survive for `.vue`, `.svelte`, `.astro`, and `.blade.php`.
- Stop-baseline capture stays on for those files.
- The page-analyzer and comment-blanking gates consult the routing map instead of the last-segment regex, closing the `.blade.php` / `.html.erb` parity gap the docs asserted.
- Directory walks widen only for configured html-engine suffixes, and `vendor` is in `SKIP_DIRS`.
- The Cursor pre-write gate resolves linked stylesheets from the real target path.
- `upsert_ext` is gone; `merge_extensions` builds a `JsMap`.

On the oracle: the twenty goldens you listed replay clean again. Eight new cases are recorded and reviewed by hand, and `DELTAS.md` carries the reviewed list instead of a pending note. `detect-help` is re-recorded for the two intended contract additions.

Two I did not touch: `live-accept-svelte-component-accept` still differs, and `live-accept-svelte-component-styled-variant` crashes in its own setup reading `node_modules/.impeccable-live/<id>/v2.svelte`. Both reproduce identically on the pre-fix head, so I left them rather than record a pre-existing difference into a golden.

mvanhorn:
Rebased onto main and addressed the bot findings in 7349565: the incomplete_styles gate is now per element (plus ancestors) rather than document-wide, so an unresolved class or a lang="scss" block no longer switches off skipped-heading and contrast findings elsewhere; configured markup suffixes such as .html.erb now get their HTML comments blanked before matching, same as .vue and .html; and the blade co-scan test was retargeted to expand_scan_targets after the rebase. cargo test --workspace is green locally.

mvanhorn:
Thanks for the depth here — the reproductions made this straightforward to act on. Reworked along the direction you suggested; the routing map stays but markup-bearing non-`.html` files now run the text pipeline through the HTML adapter and merge, so the text-rule family no longer disappears for SFCs.

Point by point:

- **Text-rule family / Stop-baseline:** both fixed. Baseline capture and both reconcile calls are restored for SFC-routed files, so the backlog-dump-at-Stop you reproduced with `Card.vue` is gone.
- **Cache key shape:** session cache is now version 2; v1 entries keyed `rule:0:snippet` are discarded rather than missed by `dedupe_against_cache`.
- **Scope:** `vendor` is in `SKIP_DIRS` and an `engine: text` configured suffix no longer widens the walk into dependency trees.
- **Sixth extension list:** replaced with one `is_scannable(path, &config.extensions)` in `impeccable-detect`, used by the walker and the hook paths.
- **Page-analyzer gate:** now consults the suffix map instead of the last-segment regex, so the `.blade.php` / `.html.erb` parity CLI-CONTRACT.md asserts is actually true.
- **`upsert_ext`:** gone in favour of `JsMap`.
- **Test gate:** goldens re-recorded and reviewed, `DELTAS.md` entry added, plus a new `hook-template-baseline-html-erb` case. `bun run test` exits 0, including `detect-dir-json-all-fixtures` matching its golden rather than riding the accepted-delta list.

`file_system.rs`, `html/src/engine.rs` and `before_edit.rs` already satisfied their findings at the current head, so I left them untouched. `cargo fmt --check` and `clippy -D warnings` still fail repo-wide on crates this branch doesn't touch.

PR #808, opened 2026-09-11
fix(detect): broken-image ignores an <img> written in prose

master5d:
## What

Both `broken-image` matchers now ignore an `` that is *written about* rather than *shipped*.

## Why

The matchers run line by line, so the regex `])*>` matches the literal text `` wherever it appears — including inside a comment or an inline-code span.

On one project this produced **seven findings in a row across three files, none of them real**. Every hit was a doc comment explaining a component:

```
 * the two rail `<img>`s get their `--px-y` written directly
```

```css
/* writes `--px-x` straight onto each `<img>`, opposite sign per side */
```

Meanwhile the built page carried only resolved sources — verified in a browser: fifteen distinct image URLs, zero empty or placeholder. The rule was, in effect, penalising documentation: the more carefully the components were commented, the more findings it raised.

## How

Two line-local guards, since a line is all a line matcher sees:

- **`is_prose_at(line, index)`** — the match sits after `//`, `/*`, `<!--` or `#`, on a doc-comment continuation line starting with `*`, or inside an unclosed backtick span.
- **`is_bare_img_tag(tag)`** — `` or `` with no attributes at all. Shipped markup always carries something (`alt`, `class`, `loading`); a bare tag is how a person writes the element's name in a sentence.

Deliberately *not* done: masking comments globally before the matchers run. That would change behaviour for every rule at once; this change is scoped to the rule that was wrong.

## Tests

Added to `regex_matchers::tests::matchers`, covering both directions:

| input | expected |
|---|---|
| `` * the rail ``s get their offset `` | silent |
| `// an  here is prose, not markup` | silent |
| `<!-- an  mentioned in a comment -->` | silent |
| `` (lagging) writes onto each `` `` (backtick span) | silent |
| `` and `` | silent |
| `Logo` | **caught** |
| `hero` | **caught** |
| `placeholder` | **caught** |

`cargo test -p impeccable-detect` → 31 passed, 0 failed. `cargo fmt --check` on the touched file is unchanged from the base branch (the file already had six pre-existing diffs; my lines add none).

## Note

The same fix applies to the JS engine at `skills/impeccable/scripts/detector/engines/regex/detect-text.mjs`, which still ships the original matchers in the packaged plugin (4.0.2 in my cache). I patched that copy locally to unblock myself and verified the same eight cases there. Happy to include the JS side in this PR if the two engines are still both live — say the word and I will push it.

---

> [!NOTE]
> **Low Risk**
> Scoped to broken-image matcher `test` logic and new helpers; real broken-image cases remain covered by added tests.
> 
> **Overview**
> Reduces false positives on the **`broken-image`** rule when source lines mention `` in documentation rather than shipping broken markup.
> 
> Both line matchers now filter matches with **`is_prose_at`**, which treats hits after comment openers (`//`, `/*`, `<!--`, `#`), on `*` doc-comment continuation lines, or inside an unclosed backtick span as prose. The matcher for tags missing `src` also skips **`is_bare_img_tag`** (``, ``, ``) so bare element names in text are not flagged, while attributed tags with empty or placeholder `src` and real missing-`src` markup still report.
> 
> Unit tests in `regex_matchers::tests::matchers` cover prose, bare tags, and unchanged real-defect cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 139e97643044910966e1841468be75d446fda7d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/detect/src/regex_matchers.rs
+++ b/crates/detect/src/regex_matchers.rs
@@ -775,6 +775,45 @@ fn transition_fmt(prefix: &str, m: &MatchCtx) -> String {
     }
 }
 
+/// `true` when the match at `index` sits in prose rather than in markup: after
+/// a line comment opener, on a doc-comment continuation line, or inside an
+/// inline-code span. Matchers see one line at a time, so these are line-local
+/// checks — enough for the case that matters, a comment that talks about an
+/// element by name.
+fn is_prose_at(line: &str, index: usize) -> bool {
+    let before = match line.get(..index) {
+        Some(b) => b,
+        None => return false,
+    };
+    for opener in ["//", "/*", "<!--", "#"] {
+        if let Some(at) = before.find(opener) {
+            let starts_token = at == 0
+                || before[..at]
+                    .chars()
+                    .next_back()
+                    .map(char::is_whitespace)
+                    .unwrap_or(true);
+            if starts_token {
+                return true;
+            }
+        }
+    }
+    if line.trim_start().starts_with('*') {
+        return true;
+    }
+    before.matches('`').count() % 2 == 1
+}
+
+/// `true` for `<img>` / `<img />` — a tag with no attributes at all. Shipped
+/// markup always carries something (`alt`, `class`, `loading`); a bare tag is
+/// how a person writes the element's NAME inside a sentence.
+fn is_bare_img_tag(tag: &str) -> bool {
+    let t = tag.trim();
+    t.eq_ignore_ascii_case("<img>")
+        || t.eq_ignore_ascii_case("<img/>")
+        || t.eq_ignore_ascii_case("<img />")
+}
+
 /// Hand-written port of `/<img\b(?:(?!\bsrc\s*=)[^>])*>/gi`.
 fn find_img_without_src(line: &str) -> Vec<MatchCtx> {
     let mut out = Vec::new();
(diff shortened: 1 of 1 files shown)


master5d:
Closing this myself — I was fixing a bug that the current code no longer has, and one of my two guards would have removed a legitimate detection. Details, since the investigation may still be useful:

**The false positives I saw came from the JS engine, not from this code.** My cached plugin was 4.0.2, whose `skills/impeccable/scripts/detector/engines/regex/detect-text.mjs` runs the matchers over raw source. There, a doc comment naming `` matched. After updating to 4.3.1 I re-ran the same three files through the shipped binary: zero findings, while the three real defects (`src=""`, `src="#"`, attributes but no `src`) were still caught.

**Why the Rust path is already correct:** `detect_text.rs` strips comments before the matchers ever see a line (`strip_js_comments` / `strip_css_comments` / `strip_css_in_js_comments` at the `run_regex_matchers` call site). That is the general fix, applied once for every rule — better than the per-rule guard I proposed. I confirmed it by building this branch's base commit and running it against the same prose files: zero findings without my patch.

**And my second guard was wrong.** `is_bare_img_tag` would have silenced `` and ``. Those are genuinely broken markup, and your binary flags them today — I verified on a plain HTML file with a bare tag and got the two findings I would have suppressed. Merging my change would have made the rule worse.

Sorry for the noise, and thanks for the comment-stripping design — it is the right shape.

PR #811, opened 2026-09-11
Add human component review and prepare Impeccable 4.4.0

pbakaus:
Comp-led builds need a user checkpoint between producing components and assembling the page. This adds Impeccable’s own component review UI and native capture/approval workflow, carries forward the comp fidelity and completion fixes, and prepares skill **4.4.0** with engine **0.1.6**.

- Review actual raster assets and rendered HTML/CSS/SVG against their regions in the approved comp. Request repairs or mark missing regions; unchanged approvals carry between rounds. Review the assembled first viewport before completion.
- Capture original PNG/WebP/JPEG bytes and rendered code outside the builder project. Bind the measured inventory centrally; every measured region must be represented, and text/controls must have code previews. Pending, needs-work, stale dependencies and a different manifest cannot pass `component-review verify`. Static previews are supported; unsupported scripted/canvas components stop rather than silently flattening or disappearing.
- Isolate code components by explicit DOM targets (manifest schema 2), excluding independently reviewed children and raster placements. Show the original assembled crop through **In context**, with a single decision for the isolated component. Preserve transparency and verify authored layout/styles; reject ambiguous or missing targets. Existing review records remain readable and are labeled as legacy region captures.
- Present the initial component kit before gate-driven automatic asset repairs so the user can redirect unsuitable production choices early.
- Retain native plate/hero capture, semantic-control evidence, stale-evidence rejection and final completion checks. Human approval does not waive those integrity checks. Embedding harnesses own human authentication and durable provider continuation.
- Bump skill manifests and generated provider distributions to 4.4.0; workspace engine, launcher pin and optional platform pins to 0.1.6. See `docs/RELEASE-4.4.0.md` for release order.

Human-approved assembled pages provide a source-bound reference for disputed text styling at native finish. Original comp scores remain unchanged; only text regions that still match the approved rendering can resolve that specific contradiction. Missing regions, controls, overall fidelity, raster integrity, stale inputs and native capture checks remain blocking. The report records the review revision and affected text regions.

Validation: Rust workspace tests passed; the default suite passed **418 Bun + 542 Node tests**; shared UI/model/viewport/bundle checks passed; native approval verification and the private harness’s real raster+code capture/repair-handoff test passed. New-work browser E2E passed (20 tests). The live-mode engine sweep passed (36 tests, 1 skip); provider-billed cleanup skipped for missing DeepSeek credentials, and the DeepSeek adapter sweep remains unverified locally. These must not be inferred from offline coverage. CI’s first bundle-byte mismatch was diagnosed as Bun version drift; `.bun-version` now pins the build toolchain without weakening the drift check. Windows browser discovery now handles case-preserved environment names. Full Rust/default suites passed again after the image/inventory changes. A real branch eval reached a durable human pause with 17 captured components (16 rendered code plus one WebP image); no human approval or completed-sample claim is implied.

Component-isolation validation: native Chromium pixel regressions cover nested ownership, raster placements, pseudo-elements, transparency, context fidelity and invalid selectors. A separate copy of the 17-component hotel kit captured successfully and was inspected in the shared UI. Existing waiting runs and human feedback were not rewritten.

**Release candidate only:** engine 0.1.6 and its platform packages must be published before skill 4.4.0; refresh the optional-package lock resolutions after publication. No engine, skill or website publication is performed by this PR update. Branch evals use the explicitly built and frozen binary, never an older downloadable engine.

Shared region authoring now includes `comp-spec --inspect-map`: numbered overlays, exact original crops, foreground-mask previews, consolidated geometry errors and explicit container/group relationships, before asset generation. It writes reference-only artifacts, leaves specs/build state/approvals unchanged, and retains every group member and raster review. The hotel replay identifies nine fully masked photo references in the recorded overlapping map; the later map has no hard geometry errors, without implying visual acceptance. Validation for this addition: 572 Rust tests, release/default Bun–Node suites and skill build passed; browser inspection verified images and layout. No grading thresholds were relaxed.

AI assistance: implemented and validated with OpenAI Codex. Visual judgments and component approvals remain human decisions.

---

> [!NOTE]
> **Low Risk**
> Changes are skill documentation, version pins, and agent workflow rules—not runtime application code—though comp-led builds become stricter about human approval gates.
> 
> **Overview**
> Releases **Impeccable 4.4.0** (skill manifests, Claude/Cursor/agent copies, plugin metadata) and bumps the bundled CLI pin to **0.1.6**, with **`.bun-version`** set to **1.3.13**.
> 
> Comp-led builds gain a **human component-review** step: new `component-review.md` defines manifests under `.impeccable/review/components.json` (and a later `hero.json`), capture/serve/verify via the `impeccable component-review` CLI or a harness `component_review` tool. Users approve real rasters and rendered code against the comp before page assembly; asset producers must hand off to this checkpoint instead of self-approving.
> 
> The **new-work** comp-led pipeline now requires `build-phase start … --artifact <entry>`, runs component review after the initial plate batch (before relying only on automatic gate repairs), documents richer hero `comp-diff` reporting, and requires `build-phase finish --disposition …` before shipping. The finish reviewer’s **Persistence** check expects every pre-`review` build phase to be closed or skipped, not only comps/spec/plates/hero.
> 
> Claude project **SessionStart** runs the existing `impeccable hook` for session prep alongside PostToolUse UI checks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37754ce3e1d9b1c77edd5b5d43209190c85c4b7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Native review now carries submitted visual approvals across shared-source edits
when verified captured images and review scope remain identical. Source revisions
and integrity checks are retained. Changed presentations still require review;
existing unsubmitted drafts can be refreshed without rewriting packet identities
or previous receipts. Implemented and tested with AI assistance.

Production-plate validation now detects comp pixels after metadata stripping,
resampling, mild blur, and small translations. It checks raw and prepared
references, preserves the texture-patch exception, and rejects invented prompt
metadata as a way to evade the copy check. Synthetic regressions cover hard
edges and false-positive cases; an isolated replay of the Gemini hotel run
rejects all five observed room-photo replacements. This remains a targeted
copy detector, not a universal proof of image provenance.

Validation for 026e340d3: full Rust workspace tests, rebuilt release engine, and
the default Bun/Node suites including oracle replay passed. No visual scoring
thresholds or builder prompts changed. Implemented and verified with AI assistance.

Region remeasurement now retains existing typography only for unchanged text/control regions against identical decoded reference pixels. Changed text, bounds, or comp pixels invalidate the measurements; legacy specs remeasure once. Component-review validation reports the offending ID and normalized bounds. Regression tests reproduce both previous failures. Full Rust workspace, rebuilt release engine, provider build, and default Bun/Node suites pass at 711066550. The hand-reviewed oracle delta adds only the comp fingerprint field; no gates or scoring thresholds change. AI-assisted implementation and validation.


pbakaus (commented):
## Adversarial review at `b9f06f3c7`: not ready to merge yet

The core design is in good shape. The local review server is secure, the capture/CDP plumbing is careful, and `cargo test --workspace` passes (every crate, zero failures). What blocks the merge is release ordering plus one gate hole that defeats the point of the feature; two independent review passes each found that hole.

### Blockers: release and process

1. **Engine 0.1.6 is not published.** There is no `engine-v0.1.6` release and no `@impeccable/cli-*@0.1.6` packages. Because the PR bumps `ENGINE_VERSION`, merging now would let the sync workflow ship provider dirs whose launcher downloads a binary that does not exist, so every first run would fail. The order is engine release first, then merge.
2. **The branch conflicts with main.** It is 57 commits behind and missing the 4.3.1 release. The conflicts are in the version manifests, `detect-antipatterns-browser.js` and the detect goldens. CI has therefore never run on the head; only Greptile and Bugbot reported.
3. **The committed provider copies are stale and contradict the source.** The tracked `reference/component-review.md` still has `schemaVersion: 1` and the old one-file-per-component preview layout. `region-map.md` is missing from every provider dir, although `new-work.md` links to it. Either drop the generated output from this PR, per policy, or regenerate it with `build:release`.

### Blockers: correctness

4. **One accepted hero auto-approves every later build in the project.** Two separate passes CONFIRMED this.
   - Where: `context/src/component_review/verify.rs:8-10`, `lifecycle.rs:44-46, 79-92`, `store.rs:86`.
   - Cause: sessions are keyed only by the project path, and `verify` returns the old receipt before it reads the manifest.
   - Repro: with one closed hero session in the store, `component-review verify --manifest .impeccable/review/components.json` (the file does not exist) exits 0 with the old approved receipt, and `lifecycle --require components` returns `accepted`.
   - Effect: a new comp build months later passes human review with no human involved. The approval needs to be bound to the build (`startedAt` or the comp hash).
5. **Files with a BOM or non-UTF-8 text always fail capture.** CONFIRMED against real Chrome.
   - Where: `browser/src/response_capture.rs:~167`, which stores `body.as_bytes()`.
   - Cause: Chrome hands back decoded text bodies, and those are compared byte-for-byte with the frozen file in `cli/src/capture_snapshot.rs:211` and `cli/src/component_capture.rs:114`.
   - Repro: an HTML file with a UTF-8 BOM came back as 285 bytes against 288 on disk; a CSS file with one Latin-1 byte came back as 30 against 28.
   - Effect: any Windows-edited project hard-fails the native build gate and component review. Capture the raw bytes or read the file directly.
6. **The font probe rejects normal system font stacks.** CONFIRMED.
   - Where: `cli/src/component_fonts.js:16-24`, `component_capture.rs:135`.
   - Cause: only the first family is tested, and it is JSON-quoted, so keywords like `-apple-system` never match.
   - Repro: `font: -apple-system, BlinkMacSystemFont, sans-serif` fails on macOS itself. `'Segoe UI', sans-serif` fails on macOS, and Georgia would fail on Linux.
   - It also scans the whole document before isolation, so an unrelated font elsewhere in the kit fails the component.
7. **Unattended runs block forever.**
   - `skill/reference/component-review.md:59` says to start `serve` in the background and wait for the user. Unlike new-work's exit-4 path, it has no `IMPECCABLE_QUESTION_DISABLED` or CI escape.
   - `server.rs:92` loops with no idle timeout, no parent watch and no teardown, and `service.json` is never removed.
   - The reviewing machine had 21 orphaned `component-review serve` processes from eval runs, some from Sept 12.

### Should fix before release

- **comp-spec accepts zero, negative and out-of-frame `box` regions.** CONFIRMED. The normalized-box branch in `comp_spec.rs:~470` has no bounds check, while pixelBox and map_inspection do. Combined with the new 1px sampling in `comp_diff.rs:287`, such a region almost always matches.
- **A band-only `--auto` draft passes the spec gate once `"draft"` is deleted.** CONFIRMED. Band regions skip the note check, and `gate_spec` only looks at `spec.draft`.
- **One corrupt session anywhere in `~/.impeccable/component-reviews` breaks `lifecycle` and `verify` for every project.** CONFIRMED. The cause is `store::read(...)?` in `lifecycle.rs:79`; skip unreadable sessions instead.
- **`verify` trusts stored booleans.** CONFIRMED.
  - Setting `capture.schema` and `receipt.captureVerified=true` in `current.json` passes `verify`, as does `components: []`.
  - A POST to `/decision` with spoofed `Origin` and `Sec-Fetch-Site` headers from curl records a browser approval.
  - No local scheme fully stops an agent that has shell access. `verify` should still recompute the capture hashes and check that `capture.components` covers the manifest. Note that `tests.rs:21` forges the capture the same way.
- **The Gemini `BeforeTool` hook prefixes every `run_shell_command` with `export IMPECCABLE_SESSION_ID=...`.**
  - Where: `hook/src/build_completion.rs:16-24`. This happens in every project, whether or not a build is running.
  - Effect: allowlists keyed on the first word of a command may re-prompt or block.
  - The hook command uses POSIX `[ ! -f ... ]` on Windows Gemini too (`skills/src/hook_manifest.rs:165`).
  - A Gemini `settings.json` with comments makes install fail, and the suggested `--force` replaces the whole file.
  - `hooks on` / `hooks reset` don't know about `.gemini`.
- **The italic-serif detector change (`b9f06f3c7`) is out of scope.** I'd split it into its own PR.
  - It reverses a deliberate "roman h1 with italic em" pass case.
  - It changes 8 detect goldens with no `DELTAS.md` entry, and it needs an extension version bump.
  - Likely false positive: `<h1>Review of <cite>Dune</cite></h1>` set large in a serif face.
- **A forced plates phase makes the next plates recheck fail.** PLAUSIBLE, not reproduced. `advance` stores forced plates as `invalid`, so hero and responsive must be forced again, and a native `finish --disposition ship` has no force path at all (`build_phase.rs:2331`, `revalidate_plates` at 664).
- **The Stop-hook completion reminder has three gaps.**
  - It fires in `ios` projects, because it runs before the native-platform check (`hook.rs:644`).
  - It pushes the model to continue exactly when component review says to wait.
  - A build started without `--artifact` never gets a reminder, and its status stays "unverified".
- **The SessionStart handler ignores the hooks-disabled setting.** It writes `IMPECCABLE_SESSION_ID` to `CLAUDE_ENV_FILE` without checking `read_config(...).enabled`.

### Minor (follow-up is fine)

- **Store lock.** The stale-lock takeover in `store.rs:45-53` can race, and `Lock::drop` deletes a lock another process now holds. The lock is also held across the whole browser capture, so a human submit during a capture gets "review is busy".
- **Pseudo-elements escape the crop guard.** `component_isolation.js:43-51` ignores overflowing `::before`/`::after`/`::marker`. CONFIRMED: a 50×50 target with a 200px `::after` reported 50×50.
- **New Chrome tests fail when Chrome won't launch.** They `unwrap()` the launch instead of skipping, as `csp_passivity.rs` does (`browser/tests/isolated_world.rs:11`, `response_capture.rs:52,127`).
- **Tight text crops lose their colour finding.** The edge filter in `hero.rs:124-137` drops real ink.
- **Review groups share one approval when they match only on `kind` and `container`** (`review_groups.rs:20`). The `reviewGroup` length is also checked in bytes while the schema counts characters.
- **UI state leaks.**
  - Undo leaves `applyGroup` behind (`review.ts:300`, `:315`).
  - A cancelled comparison-close animation leaves `closingComparison` set, which locks decisions until reload (`review.ts:407`).
- **Capture service robustness** (`capture_service.rs`).
  - A slow local client can stall it before the capability check.
  - It exits on EMFILE.
  - The ready file is written non-atomically.
  - `capture-server` is missing from `CLI-CONTRACT.md`.
- **Unexplained golden changes.** `build-phase-start-status` and `build-phase-usage` changed with no `DELTAS.md` entry.
- **`test:skill-behavior` is owed.** `new-work.md` changed, and the run isn't recorded.

### Verified as sound

- **Existing failure.** The `live-server-leak.test.mjs` SIGKILL failure is pre-existing: it fails the same way on main with the 0.1.5 binary.
- **Build gates.** `bun run build` (prose, skill prose, counts, plugin versions), `cargo xtask bundle --check` and `bun install --frozen-lockfile` all pass. The UI bundle matches `ui/`, and the UI tests pass 21/21.
- **Command count.** component-review is a playbook reached from new-work, not a 24th command.
- **Review server security.**
  - It binds `127.0.0.1` on an ephemeral port.
  - A foreign `Host` gets 403, so DNS rebinding fails, and a foreign `Origin` gets 403.
  - Encoded `..` paths are rejected, and files are served only through the hash map with digest checks.
  - Captured HTML goes into a sandboxed iframe under a `script-src 'none'` CSP.
- **Capture internals.** CDP id and event handling, isolated-world binding, path confinement, profile reservation and Chrome teardown on launch errors are all correct. The `discovery.rs` change only adds candidates.
- **Bot flags that don't hold.**
  - "Late resources escape verification" fails closed, because `revision` is checked at `component_capture.rs:163`.
  - The "1 need work" test flag is wrong; the code renders "needs" for one.
  - The control colour change in `build_phase.rs:1069` is additive.

### Suggested path

1. Publish engine 0.1.6 and its platform packages.
2. Split out the italic detector commit.
3. Fix blockers 4 to 7, plus the region validation and the corrupt-session skip.
4. Rebase onto main.
5. Drop or regenerate the provider output.
6. Let CI run.

---
AI assistance disclosure: this review was produced by Claude Code (Claude Opus 5.5), working under @pbakaus's instructions. Findings marked CONFIRMED were reproduced by running the code; the rest come from reading it.

